PR #19, opened 2021-10-13
Remove IStringMethods.cat

wenleix:
Summary:
Support "+" operator on String column -- which is most useful.

We may want to support `cat` method with simplified
API in the future.

Differential Revision: D31601427


wenleix:
As @OswinC  pointed out, we can simulate `cat` with `+` fairly straightforward. Here is the example code: 

```python
def cat(cols: List[IStringColumn], sep: str = ""):
  to_join = [cols[0]]
  for i in range(1, len(cols)):
    to_join.append(sep)
    to_join.append(col[i])
  return reduce(operator.add, to_join)
```

PR #44, opened 2021-10-30
add __contains__(), get_default() to DataFrame

wenleix:
Summary:
as title. it will make life easier

get() is registered by icolumn for index fetching... we can generalize it as well if preferred

Differential Revision: D32032534


wenleix:
Landed as https://github.com/facebookresearch/torcharrow/commit/c95a9f766d79ffd0f18c9f9c4b03c925ca618aea

PR #45, opened 2021-10-30
Some API cleanup

wenleix:
Summary:
* Rename `IColumn.isnullable` -> `is_nullable`, as PyArrow API convention is snake case. Mark with EXPERIMENTAL API
* Make `IColumn.length` as property
* Mkae `IColumn.null_count` as property -- consistent with PyArrow
* Move most getter methods to private: `getmask`, `getdata`, `get`, `gets`, `slice`, `items`, `data` -- we are having too many of them and it's quite confusing, and the `__get_item__`  seems to cover all use case (together wit the fact that `IColumn` itself is iterable.
* Rename `IColumn.valid(idx)` to `IColumn.is_valid_at(idx)`. So PyArrow has `is_valid` and `is_null`, but it returns an array indicates each element's validity/nullability. Looks like `is_valid_at(idx)` is the right way to go. Mark it with EXPERIMENTAL API.
* Mark `IColumn.tail()` as EXPERIMENTAL API, since it's not streaming-friendly (and `.head()` should be preferred to use
* Move `IColumn.count()` as private aggregate function. It's confusing to have `count`, `length` and `null_count`.

--------------------------

Note on aggregation function:

In general, the purpose for all these aggregation functions (`min`/`max`/`all`/`any`/`sum`/`prod`/`mean`/`median`/`mode`/`count`) is to to support `groupby` (See https://github.com/facebookresearch/torcharrow/blob/861609fdb5b05a6e2b2952f1ef883f3d2dda217d/torcharrow/velox_rt/dataframe_cpu.py#L2064-L2115, it finally delegates to `IColumn.AGGR_FUNC()`: https://github.com/facebookresearch/torcharrow/blob/861609fdb5b05a6e2b2952f1ef883f3d2dda217d/torcharrow/dtypes.py#L767-L778

TorchArrow indeed intends to support `groupby`, especially streaming aggregation as it's useful in feature preprocessing. However, how to support it is an open question. Especially we also want to support UDAFs. Plus, these global aggregation function is not streaming-friendly. So will move them as private in separate PRs.

Differential Revision: D32041712

### DIFF
--- a/torcharrow/dtypes.py
+++ b/torcharrow/dtypes.py
@@ -774,7 +774,7 @@ _agg_ops = {
     "mean": lambda c: c.mean(),
     "median": lambda c: c.median(),
     "mode": lambda c: c.mode(),
-    "count": lambda c: c.count(),
+    "count": lambda c: c._count(),
 }
 
 
--- a/torcharrow/icolumn.py
+++ b/torcharrow/icolumn.py
@@ -114,9 +114,25 @@ class IColumn(ty.Sized, ty.Iterable, abc.ABC):
 
     @property  # type: ignore
     @traceproperty
-    def isnullable(self):
-        """A boolean indicating whether column/frame can have nulls"""
+    def is_nullable(self):
+        """
+        EXPERIMENTAL API
+
+        A boolean indicating whether column/frame can have nulls
+        """
         return self.dtype.nullable
+
+    @property  # type: ignore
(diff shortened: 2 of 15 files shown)


wenleix:
Landed as https://github.com/facebookresearch/torcharrow/commit/a5a4da5308bbaa3465c251277306e0469c902052

PR #47, opened 2021-11-01
Minor cleanup to "cleaning APIs" and "batching APIs"

wenleix:
Summary:
* Mark `batching/unbatching` as experimental API
* fillna -> fill_null
* dropna -> drop_null

To be consistent with PyArrow API (plus today's behavior only drops null but not nan).

Also note the semantic of `df.drop_duplicates(subset)` is similar to the following SQL aggregation:
```SQL
SELECT col1, col2, ARBITRARY(col3), ARBITRARY(col4)
FROM ...
GROUP BY col1, col2
```
here `col1`, `col2` are in `subset`; `col3`, `col4` are the rest columns.

Differential Revision: D32051423

### DIFF
--- a/torcharrow/icolumn.py
+++ b/torcharrow/icolumn.py
@@ -818,9 +818,9 @@ class IColumn(ty.Sized, ty.Iterable, abc.ABC):
 
     @trace
     @expression
-    def nunique(self, dropna=True):
+    def nunique(self, drop_null=True):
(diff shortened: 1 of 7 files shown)


dracifer (approved):

wenleix:
Landed as https://github.com/facebookresearch/torcharrow/commit/b9638e6b5c262081b3cb98e0ade663e33b34605a

PR #50, opened 2021-11-03
Add docstring and tests for DataFrame construction from tuple data

wenleix:
Summary:
In addition to using `dtype` we can also use `columns` to inform the column names when constructing DataFrame from tuple data
```
ta.DataFrame([(1,'a'), (2,'b')], columns=['t1', 't2'])
```

Differential Revision: D32142919

### DIFF
--- a/torcharrow/idataframe.py
+++ b/torcharrow/idataframe.py
@@ -56,6 +56,13 @@ def DataFrame(
         where possible.  Should be a dt.Struct() providing a list of
         dt.Fields.
 
+    columns: list of strings, default None
+        The name of columns. Used when data is a list of tuples without a custom
+        dtype provided. len(columns) should be equal to len(data[0]). When data
+        is a list of tuples and dtype is provided this will be ignored.
+        This should be left to be None when data and dtype are both None (the
+        semantic is constructing a default empty DataFrame without any columns).
+
     device: Device, default ""
         Device selects which runtime to use from scope.  TorchArrow supports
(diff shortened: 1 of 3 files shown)


torcharrow/idataframe.py
```diff
@@ -56,6 +56,14 @@ def DataFrame(
         where possible.  Should be a dt.Struct() providing a list of
         dt.Fields.
 
+    columns: list of strings, default None
```
wenleix, line 59: list of strings is good for now. Note in the future we might want to support nested column names. But how to specify it is still open. We can use a dict but it looks yunky for non-struct type: 
```python
ta.DataFrame(
    [(1, 2, (3, 4))], 
    columns={
           "a": None, 
           "b": None, 
           "nest": ["n1", "n2"]
   }
)
```

Or we can have a recursive type `FieldDesc`:
```python
class FieldDesc:
    fieldName: str
    nested_fields: Optional[List[Union[String, FieldDesc]]]
```

```python
ta.DataFrame(
    [(1, 2, (3, 4))], 
    columns=["a", "b", FieldDesc("nest", ["n1", "n2"])]
)
```

wenleix (commented):
LGTM.

torcharrow/idataframe.py
OswinC: Does allowing only NamedTuple (no "pure" tuple) sound like an option?

PR #52, opened 2021-11-04
Delegate IColumn.min/max to Arrow Compute min/max

wenleix:
Differential Revision: D32164033

### DIFF
--- a/torcharrow/velox_rt/numerical_column_cpu.py
+++ b/torcharrow/velox_rt/numerical_column_cpu.py
@@ -713,30 +713,6 @@ class NumericalColumnCpu(INumericalColumn, ColumnFromVelox):
 
     @trace
     @expression
-    def min(self, numeric_only=None, fill_value=None):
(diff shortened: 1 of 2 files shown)


torcharrow/velox_rt/numerical_column_cpu.py
```diff
@@ -895,25 +871,6 @@ def is_monotonic_decreasing(self):
...
-    # interop ----------------------------------------------------------------
-
-    @trace
-    def to_pandas(self):
-        """Convert self to pandas dataframe"""
-        # TODO Add type translation
-        # Skipping analyzing 'pandas': found module but no type hints or library stubs
-        import pandas as pd  # type: ignore
-
-        return pd.Series(self._ma())
```
wenleix, line 907: `._ma()` doesn't exists in cpu column (it's the numpy masked array). So this causes `to_arrow/to_pandas` failing on cpu column. 

Remove them for now so it use the default `to_array()` implementation by Python iterable.

PR #63, opened 2021-11-13
Basic PyTorch interop

wenleix:
Summary:
It's a prototype implementaiton to demonstrate the semantic. It's not efficient and a bit complicated, but a good initial step.

Tests are from https://github.com/facebookresearch/torcharrow/blob/9927a9c428c99ec8c4be3352c318ed2c66d36990/torcharrow/test/test_legacy_interop.py

Implementation is based on the demo backend (Removed in https://github.com/facebookresearch/torcharrow/commit/8bc63029fd555a04071834f605bb67de31baa589). But use arrow format as intermediate type.

Differential Revision: D32407416

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -57,7 +57,9 @@ jobs:
           CCACHE_DIR=$GITHUB_WORKSPACE/.ccache DEBUG=1 python setup.py develop --user
 
       - name: Install dependencies for test
-        run: pip3 install pytest --user
+        run: |
+          pip3 install pytest  --user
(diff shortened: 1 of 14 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -57,7 +57,9 @@ jobs:
           CCACHE_DIR=$GITHUB_WORKSPACE/.ccache DEBUG=1 python setup.py develop --user
 
       - name: Install dependencies for test
-        run: pip3 install pytest --user
+        run: |
+          pip3 install pytest  --user
+          pip3 install torch==1.10.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html --user
```
wenleix, line 62: @ejguan  does this look right way to install pytorch in CI?

PR #66, opened 2021-11-15
Remove dep build directory from git history

ejguan:
For ubuntu and mac build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,14 @@ build/
 # OSX dir files
 .DS_Store
 
+# Dep
+_build/
+double-conversion/
+fmt/
+folly/
+googletest/
+range-v3/
+re2/
+
 # Generated Files
 torcharrow/version.py


.gitignore
```diff
@@ -12,5 +12,14 @@ build/
 # OSX dir files
 .DS_Store
 
+# Dep
+_build/
+double-conversion/
```
wenleix, line 17: hmm... we should also fix the script so they don't pollute root level ? lol

wenleix (approved):
Thanks!

.gitignore
wenleix: We can use this `DEPENDENCY_DIR` , defined in https://github.com/facebookresearch/torcharrow/blob/e03beb1750e13f96f8b8a049b6b82fb38c927841/scripts/_setup-macos.sh#L38

Here is an example usage in Velox CI: 
https://github.com/facebookincubator/velox/blob/776f9840462f7d2b5053740ffd61c5202eeeee67/.circleci/config.yml#L69

ejguan: I see. I can move them into `_build` directory.

ejguan: to align with the behavior on ubuntu

wenleix: @ejguan : Feel free to do it in a separate PR

PR #68, opened 2021-11-15
Move build directory of dep to _build

ejguan:
Only tested on my local mac

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,19 @@
 # Byte-compiled / optimized
 __pycache__/
 
+# Editor temporaries
+*.swn
+*.swo
+*.swp
+*.swm
+*~
+.vscode
+.idea
+
+# Compiled python
+*.pyc
+*.pyd
+
 # C extensions
 *.so
 
@@ -14,12 +27,6 @@ build/
 
 # Dep
 _build/
-double-conversion/
-fmt/
-folly/
-googletest/
-range-v3/
-re2/
 
 # Generated Files
 torcharrow/version.py
--- a/scripts/build_mac_dep.sh
+++ b/scripts/build_mac_dep.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 # Copyright (c) Facebook, Inc. and its affiliates.
 
-SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+ROOT_DIR="$( cd -- "$( dirname "$( dirname "${BASH_SOURCE[0]}" )" )" &> /dev/null && pwd )"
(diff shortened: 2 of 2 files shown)


scripts/_setup-macos.sh
```diff
@@ -35,7 +35,8 @@ set -x # Print commands that are executed.
 FB_OS_VERSION=v2021.05.10.00
 NPROC=$(sysctl -n hw.physicalcpu)
 COMPILER_FLAGS="-mavx2 -mfma -mavx -mf16c -masm=intel -mlzcnt"
-DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)}
+DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)}/_build
```
wenleix, line 38: Wondering if it makes sense to do something like 
```
DEPENDENCY_DIR=XXX 
```
 
in `build_mac_dep.sh` ? -- so we can keep this file sync with the Velox repo on ?

ejguan: yeah, doable.

scripts/_setup-macos.sh
```diff
@@ -35,7 +35,6 @@ set -x # Print commands that are executed.
 FB_OS_VERSION=v2021.05.10.00
 NPROC=$(sysctl -n hw.physicalcpu)
 COMPILER_FLAGS="-mavx2 -mfma -mavx -mf16c -masm=intel -mlzcnt"
-DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)}
```
wenleix, line 38: If we define `DEPENDENCY_DIR` before sourcing the `_setup-macos.sh`, do we still need to remove this line? ;)

wenleix (approved):
Thanks!

scripts/_setup-macos.sh
ejguan: Why we need to keep it as `DEPENDENCY_DIR` has been declared.

wenleix: @ejguan In theory to keep the file same as https://github.com/facebookincubator/velox/blob/78159a995038e38576c354348023f50cc49a4864/scripts/setup-macos.sh so in the future it's easy to upgrade if Velox update's the `setup-macos.sh` ?

https://github.com/facebookresearch/torcharrow/blob/3fd191043989d2130d7df93535d795bbd64c3313/scripts/_setup-macos.sh#L16-L18

ejguan: Then, why not delete this file and call the script from velox directly?

wenleix: Guess we can also do this? -- but we don't have Mac CI yet so we don't know whether our Mac build might break when update Velox submodule.

PR #70, opened 2021-11-16
Initialize default Scope in TestDFNoTrace

wenleix:
Otherwise the test run with
```
python -m unittest -v
```

would fail.

### DIFF
--- a/torcharrow/test/test_trace.py
+++ b/torcharrow/test/test_trace.py
@@ -45,6 +45,9 @@ class DF:
 
 
 class TestDFNoTrace(unittest.TestCase):
+    def setUp(self):
+        Scope.default = Scope()
+
     def test_tracing_off(self):
         trace = Scope.default.trace
 


torcharrow/test/test_trace.py
```diff
@@ -45,6 +45,9 @@ def g(self, other):
 
 
 class TestDFNoTrace(unittest.TestCase):
+    def setUp(self):
+        Scope.default = Scope()
```
ejguan, line 49: nit: If the scope is used by the whole class. You can use `setUpClass`. https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

It can also be achieved in a separate PR.

ejguan (approved):
LGTM

torcharrow/test/test_trace.py
wenleix: @ejguan : I see. The plan is  to deprecate `Scope` though: https://github.com/facebookresearch/torcharrow/pull/37

ejguan: Then, no need to worry about it.

PR #77, opened 2021-11-20
Ensure correct equality is used when testing IColumn contents.

wenleix:
Summary:
See T106347840 for details. A significant amount of our tests
were not actually testing anything and would always pass even if the
code was buggy.

Reviewed By: wenleix

Differential Revision: D32577543


wenleix:
merged via https://github.com/facebookresearch/torcharrow/commit/716e73ec3d89a53d49417763578d0473d0f35c08

PR #81, opened 2021-11-24
Use vectorized ops for string comparisons.

wenleix:
Summary: Switch string comparisons from using slow python comparators to Velox fast ones.

Differential Revision: D32630988

### DIFF
--- a/torcharrow/velox_rt/string_column_cpu.py
+++ b/torcharrow/velox_rt/string_column_cpu.py
@@ -13,6 +13,7 @@ from torcharrow.expression import expression
 from torcharrow.functional import functional
 from torcharrow.istring_column import IStringColumn, IStringMethods
 from torcharrow.scope import Scope, Device
+from torcharrow.trace import trace
 
 from .column import ColumnFromVelox
 from .typing import get_velox_type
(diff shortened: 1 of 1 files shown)


torcharrow/velox_rt/string_column_cpu.py
```diff
@@ -139,20 +140,53 @@ def __add__(self, other):
...
+            nullable = nullable or other.dtype.nullable
         else:
             assert isinstance(other, str)
-            return functional.eq(self, other).with_null(self.dtype.nullable)
+        return f(self, other).with_null(nullable)
+
+    @trace
+    @expression
+    def __eq__(self, other):
+        """Vectorized a == b."""
```
wenleix, line 158: unrelated to this diff. But i am always thinking whether we need to emphasize "vectorized"?  Here is the docstring in Pandas:

```
 |  __add__(self, other)
 |
 |  __and__(self, other)
 |
 |  __divmod__(self, other)
 |
 |  __eq__(self, other)
 |      Return self==value.
 |
 |  __floordiv__(self, other)
 |
 |  __ge__(self, other)
 |      Return self>=value.
 |
 |  __gt__(self, other)
 |      Return self>value.
 |
 |  __le__(self, other)
 |      Return self<=value.
 |
 |  __lt__(self, other)
 |      Return self<value.
 |
 |  __mod__(self, other)
 |
 |  __mul__(self, other)
 |
 |  __ne__(self, other)
 |      Return self!=value.
 |
 |  __or__(self, other)
 |
 |  __pow__(self, other)
 |
 |  __radd__(self, other)
...
```

wenleix (commented):
LGTM. Thanks for using native Velox kernels for string comparisons!

PR #90, opened 2021-11-25
Tweaking nullability setting in from_arrow

OswinC:
Summary:
Added argument `nullable` to `from_arrow` and get back the behavior of defaulting `nullable` to null_count > 0.

Removed `dtype` argument since type casting alongside with zero-copy data passing between Arrow and TA/Velox is not supported at the moment. We can always infer TA dtype from the Arrow type.

Differential Revision: D32662634


OswinC:
Abandoned

PR #93, opened 2021-11-30
Support Python 3.7

wenleix:
Google Collab runs Python 3.7


.github/workflows/ubuntu.yml
```diff
@@ -16,6 +16,11 @@ jobs:
       - name: Print CPU info
         run: cat /proc/cpuinfo
 
+      - name: Setup Python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
ejguan, line 22: Do you want to use matrix of python versions? We have a reference from torchdata. https://github.com/pytorch/data/blob/2d94ebc6e95d4bd475a98e947781e58410386a10/.github/workflows/ci.yml#L19-L28

ejguan (approved):
LGTM with a comment about python version matrix

.github/workflows/ubuntu.yml
wenleix: @ejguan we should. But the current build/test time is not not ideal so just keep the most important test for now...

setup.py
```diff
@@ -135,20 +135,21 @@ def build_extension(self, ext):
     license="BSD",
     install_requires=[
         "arrow",
-        "numpy",
+        "numpy==1.21.4",
```
wenleix, line 138: otherwise, it will pick up `1.22.0rc1` and doesn't find Python 3.7 distribution.

@ejguan  have you seen similar issues before?

Issue #64, opened 2021-11-13
Delegate `IColumn.fill_null/drop_null` to Arrow


wenleix:
Arrow array supports `fill_null/drop_null`.

So we can first call `to_arrow`, then calls `fill_null/drop_null` in Arrow array, and then converts back (`from_arrow` is working in progress)

This looks like a decent initial implementation.

PR #96, opened 2021-12-02
Delegate `IColumn.fill_null/drop_null` to Arrow

wenleix:
Summary:
As title

NOTE: the changes of diff only delegate IColumn.fill_null to Arrow since TorchArrow is using Apache Arrow 2.0.0, which doesn't support `drop_null` until Apache Arrow 6: https://arrow.apache.org/docs/python/api/compute.html#selections

Reviewed By: wenleix

Differential Revision: D32770009

### DIFF
--- a/torcharrow/icolumn.py
+++ b/torcharrow/icolumn.py
@@ -19,6 +19,7 @@ from tabulate import tabulate
 
 from .dispatcher import Device
 from .expression import expression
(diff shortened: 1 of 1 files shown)


wenleix:
Related issue: https://github.com/facebookresearch/torcharrow/issues/64

PR #104, opened 2021-12-06
Clean up imports in __init__.py

wenleix:
Summary:
with these `from XXX import *`, it's difficult to track the methods/variable get imported to the package level

Also use `__all__` in `__init__.py` for so it shows docstring for global methods in `help(torcharrow)`

Reviewed By: OswinC

Differential Revision: D32858748

### DIFF
--- a/torcharrow/__init__.py
+++ b/torcharrow/__init__.py
@@ -1,32 +1,26 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 
-# For relative imports to work in Python 3.6
-from .expression import *  # dependencies: None
-from .trace import *  # dependencies: expression
-
-# don't include
-# from .dtypes import *
-# since dtypes define Tuple and List which confuse mypy
-
-from .scope import *  # dependencies: column_factory, dtypes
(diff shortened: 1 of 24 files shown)


wenleix:
Looks like package contents still reads a bit overwhelming. Any suggestions about what's the idiomatic way to handle it in Python packages. Perhaps it's fine? Or should we move them into `interal/impl` directory? cc @VitalyFedyunin , @ejguan 

```python
import torcharrow as ta
help(ta)

...

PACKAGE CONTENTS
    _interop
    _pytorch
    _torcharrow
    dispatcher
    dtypes
    expression
    functional
    icolumn
    idataframe
    ilist_column
    imap_column
    interop
    inumerical_column
    istring_column
    pytorch
    scope
    trace
    velox_rt (package)
```

ejguan:
`__all__` would only helps what are imported when users want to do `from module import *`. IIRC, it would not change the output of `help`. If you don't want to expose it from torcharrow level, you have to move them into a subfolder like you said.

PR #114, opened 2021-12-10
Export column types and update benchmark scripts

OswinC:
Summary: Title. Some updates for following up the recent refactoring that un-exported many symbols

Differential Revision: D33019830

### DIFF
--- a/torcharrow/__init__.py
+++ b/torcharrow/__init__.py
@@ -4,7 +4,11 @@ from . import pytorch  # noqa
 from . import velox_rt  # noqa
 from .icolumn import IColumn, Column, concat, if_else  # noqa
 from .idataframe import IDataFrame, DataFrame, me  # noqa
+from .ilist_column import IListColumn  # noqa
+from .imap_column import IMapColumn  # noqa
 from .interop import from_pylist, from_arrow  # noqa
+from .inumerical_column import INumericalColumn  # noqa
+from .istring_column import IStringColumn  # noqa
 
 try:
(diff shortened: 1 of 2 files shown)


torcharrow/__init__.py
```diff
@@ -4,7 +4,11 @@
 from . import velox_rt  # noqa
 from .icolumn import IColumn, Column, concat, if_else  # noqa
 from .idataframe import IDataFrame, DataFrame, me  # noqa
+from .ilist_column import IListColumn  # noqa
```
wenleix, line 7: Unrelated to this diff, but I am also thinking to do renames like `icolumn.py -> _icolumn.py`, etc

Otherwise when I type `ta.`, a lot of internal package name (e.g.  `icolumn`, `ilist_column`) pops up in IDE hint and mixed with useful names: 

Screen Shot 2021-12-10 at 9 59 00 AM

This looks quite common practice in Python libraries. For example, in PyTorch the file name is `_torch.py`

wenleix (approved):

torcharrow/__init__.py
OswinC: Prefixing internal names with "_" sounds good to me. But names like icolumn and IColumn feel like a public name to me, e.g. it's the type of everything we return from ta

wenleix: @OswinC `IColumn` is a type that we already exposed. But `icolumn` is not the type.

Neilblaze (approved):

PR #125, opened 2021-12-22
Add packaging instruction for Linux

wenleix:
Also remove the warning about "build and install with release mode may
take very long time" -- now it's using `VELOX_BULID_MINIMAL` which reduce
the number of artifacts being built by 75%.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ And run unit tests with
 python -m unittest -v
 ```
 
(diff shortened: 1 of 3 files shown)


NivekT (commented):
Do you know the cause of the failure [here](https://github.com/facebookresearch/torcharrow/runs/4610693772?check_suite_focus=true)?

Issue #147, opened 2022-01-20
Arrow conversion for struct type 


wenleix:
Currently it's supported for numeric type

Related PR:
- to_arrow for numeric type: https://github.com/facebookresearch/torcharrow/pull/111
- from_arrow for numeric type: https://github.com/facebookresearch/torcharrow/pull/103
- DataFrame.to_arrow: https://github.com/facebookresearch/torcharrow/pull/113
- from_arrow for Arrow Table: https://github.com/facebookresearch/torcharrow/pull/110

PR #155, opened 2022-01-25
Initial support for to_arrow() for struct column

wenleix:
Summary: Once there is Velox RowVector -> Arrow StructArray conversion, we should switch to use that for better performance. But this would provide initial support.

Reviewed By: damianr99

Differential Revision: D33652891


wenleix:
Some initial support to https://github.com/facebookresearch/torcharrow/issues/147

PR #156, opened 2022-01-26
Initial support for from_arrow() for struct column

wenleix:
Summary:
Similar to D33652891 (https://github.com/facebookresearch/torcharrow/commit/8be2e4531a027c3f48340abf74a5b13a57a3b7fd) /  https://github.com/facebookresearch/torcharrow/pull/155, add basic support in Python with non-optimal perf.

Once there is Arrow StructArray -> Velox RowVector conversion, we should switch to use that for better performance

Differential Revision: D33784374


wenleix:
Some initial support to #147

PR #165, opened 2022-02-01
Fix Modulus and update to generate expr eagerly

vancexu:
Summary:
Problem:
TorchArrow mod operator works different from PyTorch/Pandas right now, because C++ and Python implement mod differently. Basically
- in python, result of a % b is: a - floor(a / b)
- in c++, result of a % b is: a - int(a / b)

This diff address it with a new velox udf floormod

Differential Revision: D33924566

### DIFF
--- a/csrc/velox/column.cpp
+++ b/csrc/velox/column.cpp
@@ -467,7 +467,7 @@ std::string opCodeToFunctionName(BinaryOpCode opCode) {
       return "multiply";
     } break;
     case BinaryOpCode::Modulus: {
-      return "mod";
+      return "torcharrow_floormod";
     } break;
     case BinaryOpCode::Eq: {
       return "eq";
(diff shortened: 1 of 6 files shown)


vancexu:
@wenleix Would you mind help a look into the "Ubuntu Build and Test failure"? Is this a real issue and should block landing internal diff? 
How do you usually debugging this? I tried clone to local macbook but pytest produce different error even in main branch. (my local error is `ModuleNotFoundError: No module named 'torcharrow._torcharrow'`)

vancexu:
> @wenleix Would you mind help a look into the "Ubuntu Build and Test failure"? Is this a real issue and should block landing internal diff? How do you usually debugging this? I tried clone to local macbook but pytest produce different error even in main branch. (my local error is `ModuleNotFoundError: No module named 'torcharrow._torcharrow'`)

NVM, manually update to latest version make Ubuntu Build and Test pass. Previously it likely failed due to real test failure (even though error logs are confusing).

Issue #143, opened 2022-01-18
Naive kernel binding for cast in CPU backend


wenleix:
# Naive kernel binding for cast in CPU backend

## Background: TorchArrow Native Kernel Dispatch
For efficiency, a lot of TorchArrow operations (e.g. `INumerialColumn.abs()`) is dispatched to the Velox C++ native kernel. A Velox native kernel is represented by a compiled [Velox ExprSet](https://facebookincubator.github.io/velox/develop/expression-evaluation.html#compilation), which can be evaluated over the `RowVector` that represents input data. 

As a concrete example, here is the lifecycle how `NumerialColumnCpu.abs()` get dispatched in CPU backend: 

1. The abs operation is delegated to `NumerialColumnCpu._data.abs()`, here: 
https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/torcharrow/velox_rt/numerical_column_cpu.py#L598-L601

`NumerialColumnCpu._data` is a `SimpleColumnXXX` (e.g. `SimpleColumnBIGINT`: see [this PYI file](https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/csrc/velox/_torcharrow.pyi#L40-L48)) binded here: https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/csrc/velox/lib.cpp#L102-L103 

2. `SimpleColumnBIGINT` (and other numeric types) has `abs()` operation binded to C++ method here: https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/csrc/velox/lib.cpp#L355. 

Here is the implementation of `SimpleColumn<T>::abs`: https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/csrc/velox/column.h#L459-L465

which mainly consists of two steps:

    * Construct, compile and cache ExprSet: https://github.com/facebookresearch/torcharrow/blob/7510a2d09f6d58c8b8b7493fe6d64925ff59b0ff/csrc/velox/column.cpp#L208-L228 
Essentially it generates a CallTypedExpr, which calls the funcion abs over field c0. c0 is again represented as FieldAccessTypedExpr
[...]

PR #169, opened 2022-02-03
Add integration test for Criteo preproc

wenleix:


torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,176 @@
...
+        type(self).RAW_ROWS = rows
+        df = ta.DataFrame(rows, dtype=DTYPE)
+
+        pq.write_table(df.to_arrow(), type(self).TEMPORARY_PARQUETY_FILE)
+    
+    def tearDown(self) -> None:
+        os.remove(type(self).TEMPORARY_PARQUETY_FILE)
+
+    @staticmethod
+    def preproc(df: ta.IDataFrame) -> ta.IDataFrame:
```
wenleix, line 110: this is the structural transformation

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,176 @@
...
+                    for cat_name in DEFAULT_CAT_NAMES
+                ]
+            ),
+        ),
+    ]
+)
+
+# TODO: allow to use to_tensor with Callable
+# TODO: implement conversion in native C++
+class _DenseConversion(tap.ITorchConversion):
```
wenleix, line 48: tensor conversion prototype implemented in Python

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,176 @@
...
+
+        pq.write_table(df.to_arrow(), type(self).TEMPORARY_PARQUETY_FILE)
+    
+    def tearDown(self) -> None:
+        os.remove(type(self).TEMPORARY_PARQUETY_FILE)
+
+    @staticmethod
+    def preproc(df: ta.IDataFrame) -> ta.IDataFrame:
+        # 1. fill null values
+        df["dense_features"] = df["dense_features"].fill_null(0)
```
wenleix, line 112: `fill_null` still has O(n^2) behavior

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # 3. Pack each categorical feature as an single element array
+        sparse_features = df["sparse_features"]
+        for field in DEFAULT_CAT_NAMES:
+            sparse_features[field] = functional.array_constructor(sparse_features[field])
+
+        # FIXME: we shouldn't need to "put back the struct column"
+        df["sparse_features"] = sparse_features
+
+        df["label"] = df["label"].cast(dt.int32)
```
wenleix, line 126: cast will also be improved via https://github.com/facebookresearch/torcharrow/issues/143

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,217 @@
...
+        df["dense_features"] = df["dense_features"].fill_null(0)
+        df["sparse_features"] = df["sparse_features"].fill_null(0)
+
+        # 2. apply log(x+3) on dense features
+        df["dense_features"] = (df["dense_features"] + 3).log()
+
+        # 3. Pack each categorical feature as an single element array
+        sparse_features = df["sparse_features"]
+        for field in sparse_features.columns:
+            sparse_features[field] = functional.array_constructor(
```
wenleix, line 142: cc @VitalyFedyunin

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,217 @@
...
+# TODO: implement conversion in native C++
+class _DenseConversion(tap.ITorchConversion):
+    def to_tensor(self, df: ta.IDataFrame):
+        # Default to_tensor, each field is a Tensor
+        tensors = df.to_tensor()
+
+        # Stack them into BatchSize * NumFields Tensor
+        return torch.cat(
+            [column_tensor.values.unsqueeze(0).T for column_tensor in tensors], dim=1
+        )
```
ejguan, line 64: Is each `column_tensor` or `column_tensor.values` a real Tensor?
nit: Would `torch.cat([column_tensor.values.unsqueeze(0) for column_tensor in tensors]).T` be easier to understand?

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,217 @@
...
+        # Stack them into BatchSize * NumFields Tensor
+        packed_tensor = torch.cat(
+            [
+                column_tensor.values.values.values.unsqueeze(0).T
+                for column_tensor in tensors
+            ],
+            dim=1,
+        )
+        # Flatten the tensors into one-dimension with columnar ordering
+        values = packed_tensor.transpose(1, 0).reshape(-1)
```
ejguan, line 93: Since you are going to flatten them, why not `values = torch.cat([column_tensor.values.values.values.values for column_tensor in tensors])`?

And I'm really confused by the concept of `column_tensor.values.values.values`

torcharrow/test/integration/test_criteo.py
```diff
@@ -0,0 +1,217 @@
...
+        # FIXME: we shouldn't need to "put back the struct column"
+        df["sparse_features"] = sparse_features
+
+        df["label"] = df["label"].cast(dt.int32)
+
+        return df
+
+    def test_criteo_transform(self):
+        # Read data from Parquet file
+        table = pq.read_table(type(self).TEMPORARY_PARQUETY_FILE)
```
ejguan, line 155: nit: `self.TEMPORARY_PARQUETY_FILE` should be fine.

ejguan (approved):
LGTM with a few nit comments. And, for demo purpose, do we want to integrate it with DataLoader2?

torcharrow/test/integration/test_criteo.py
wenleix: > And I'm really confused by the concept of column_tensor.values.values.values

Yeah , this is quite confusing at this moment -- makes the default `to_tensor` difficult to use :(

The original column type is `List[Integer]`. Currently the type is incorrectly marked as "nullable" for both list and the internal integer. 

For the nullable type, it will the `to_tensor` result will be wrapped in `WithPresence`: https://github.com/facebookresearch/torcharrow/blob/4b5e8258e0e1f573a270dced7f26f3b5cf8de587/torcharrow/_pytorch/common.py#L17-L21

So there are to `.values` call required to get unwrap for `WithPresence`.

The last `.values` is to get the "flatten" array from `PackedList`, which is required: 
https://github.com/facebookresearch/torcharrow/blob/4b5e8258e0e1f573a270dced7f26f3b5cf8de587/torcharrow/_pytorch/common.py#L25-L29

Let me see if I can simplify it a bit (e.g. hack to make the column non-nullable~)

wenleix: > Since you are going to flatten them, why not values = torch.cat([column_tensor.values.values.values.values for column_tensor in tensors])?

I am copying the code from TorchRec repo ... -- `cat` by default will flatten all tensors, right?

torcharrow/test/integration/test_criteo.py
wenleix: > Is each column_tensor or column_tensor.values a real Tensor?

`column_tensor.values ` is the real Tensor -- i think i really need to hack around the nullability issue :)

torcharrow/test/integration/test_criteo.py
ejguan: It will concatenate the first dimension by default. If inputs are in shape of `(10, 1)`, the output is going to be `(20, 1)`.
And, `(10)` will be `(20)`

ejguan: > The original column type is list[int32]. Currently the type is incorrectly marked as "nullable" for both list and the internal int32.

> For the nullable type, it will the to_tensor result will be wrapped in WithPresence:

I think it's fine to be nullable. But, whenever `to_tensor` is called, it should return a Tensor. And, depending on the tensor type, if it's dense tensor, we should raise Error for null value. If it's specified as sparse tensor, we may allow null value in Tensor

wenleix:
@ejguan : 

> And, for demo purpose, do we want to integrate it with DataLoader2?

This is majorly to test end-to-end in TorchArrow (especially, the arrow parquet reader since it's currently only testable in GitHub...)

torcharrow/test/integration/test_criteo.py
wenleix: The idea is if it's a nullable int, it will return with presence(mask) Tensor by default.  The problem is `functional` now pessimistically tag everything as nullable. -- will try to fix this separately 

We do plan to have dedicated `Dense` and `Jagged Sparse` conversion . So we will remove the two collation methods anyway.

PR #92, opened 2021-11-29
Print CPU instruction set in CI

wenleix:
Summary:
We see "illegal instruction" error few times on CI (which is not reproducible).  We suspect it's due to it happened to be run a CPU without AVX support.  (https://stackoverflow.com/questions/37480071/how-to-tell-if-a-linux-machine-supports-avx-avx2-instructions)

Print the instruction set to help debug.

Differential Revision: D32707568

PR #223, opened 2022-02-28
Add support for nightly TorchArrow builds

kgpai:


kgpai:
@wenleix - Can you approve CI build ?
Unable to also add reviewers to this.

.github/workflows/velox.yml
```diff
@@ -0,0 +1,57 @@
+name: Velox Nightly Build and Test
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  vm:
+    runs-on: ubuntu-latest
+
```
ejguan, line 10: Can we remove these three lines and test it?

.github/workflows/velox.yml
```diff
@@ -0,0 +1,57 @@
...
+  container:
+    runs-on: ubuntu-latest
+    container: prestocpp/velox-torcharrow:kpai-20220228
```
ejguan, line 13: What is the container here?

kgpai: Its public velox image : https://hub.docker.com/r/prestocpp/velox-torcharrow/tags

.github/workflows/velox.yml
```diff
@@ -0,0 +1,54 @@
+name: Velox Nightly Build and Test
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  container:
+    runs-on: ubuntu-latest
```
wenleix, line 9: since there is a container specified -- would this still be useful (e.g. I imagine it's always running the linux in container? ~)

kgpai: Let me see if i can review it, i think this is supposed to be the vm in which the docker container runs..

.github/workflows/velox.yml
```diff
@@ -0,0 +1,61 @@
+name: Velox Nightly Build and Test
```
NivekT, line 1: Should this say "TorchArrow Nightly Build and Test" instead?

kgpai: Well we are in TorchArrow and we are updating velox to latest sha, thus `velox nightly build and test`

.github/workflows/velox.yml
kgpai: runs-on cant be removed its required ..

.github/workflows/velox.yml
wenleix: Yeah i would say not update the Velox to latest commit -- so we know which commits we build the TorchArrow CPython extension.

.github/workflows/velox.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Check out source repository
+        uses: actions/checkout@v2
+        with:
+            submodules: true
+
+      - name: Update submodules
+        run: |
+          git submodule update --init --recursive
+          cd csrc/velox/velox/
+          git submodule update --remote --merge
```
wenleix, line 29: can we just use the Velox commit pinned in TorchArrow? -- we can update daily, for example.

.github/workflows/velox.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ubuntu-latest
+    container: prestocpp/velox-velox-torcharrow:kpai-20220301
+    steps:
+      - name: Print CPU info
+        run: cat /proc/cpuinfo
+
+      - name: Setup Python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
wenleix, line 18: I see Python environment is set up by conda -- do we need to set up Python environment through GitHub Actions?

.github/workflows/velox.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Update submodules
+        run: |
+          git submodule update --init --recursive
+          cd csrc/velox/velox/
+          git submodule update --remote --merge
+
+      - name: Setup conda environment
+        run: | 
+          source /opt/conda/etc/profile.d/conda.sh
+          conda create --name ta_python37 python=3.7 -y
```
wenleix, line 34: @ejguan : Wondering if there is any matrix magic so we can build for Python 3.7/3.8/3.9? ^_^

wenleix:
I assume Velox submodule no longer needed to be updated? :)

.github/workflows/velox.yml
wenleix: nit: for file name, what about `torcharrow-linux-nightly.yml`?

.github/workflows/velox.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Install dependencies for test
+        run: |
+          pip3 install pytest  --user
+          pip3 install torch==1.10.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html --user
+
+      - name : Run TorchArrow unit tests
+        run: |
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib/
+          python3 -m pytest --no-header -v torcharrow/test
```
wenleix, line 50: there is chance for the test to fail with "illegal instructions", when GitHub Actions runs on a `Intel E5-2673 v3/v4` : https://github.com/facebookresearch/torcharrow/pull/92#issuecomment-982850395

.github/workflows/velox.yml
wenleix, line 34: Also when you run 
```
PYTHON_VERSION=3.7 packaging/build_wheel.sh
```

It will use conda to initialize the Python environment:

https://github.com/facebookresearch/torcharrow/blob/4d21b119bb4e3d7a9db0a9b14d9fec3d3a1b9dcd/packaging/build_wheel.sh#L45-L50

.github/workflows/velox.yml
kgpai: Sure, we can do that; I thought we wanted to build against latest Velox nightly this this change. If we dont want to do that then I will remove this.

.github/workflows/velox.yml
kgpai: I will still need conda to build TorchArrow and run unit tests right ?

.github/workflows/velox.yml
kgpai: So  are you saying we dont build torch arrow unit tests etc, and just go directly to building wheel ?

wenleix: Correct.

.github/workflows/velox.yml
wenleix: Yeah... but the published TorchArrow wheel is a bit weird -- it's not a TorchArrow nightly build. 

I guess we need Velox Python library (long term plan, similar to PyArrow) to build Velox nightly wheel.

.github/workflows/velox.yml
ejguan: Yeah. As TA has pybind, you need to have a matrix of python versions to generate corresponding nightly wheel. Reference:
- Matrix definition: https://github.com/pytorch/data/blob/823d216535782a2c31a19bcfaeb450597a53f8c4/.github/workflows/_build_test_upload.yml#L62-L70
- Reference python version in steps: https://github.com/pytorch/data/blob/823d216535782a2c31a19bcfaeb450597a53f8c4/.github/workflows/_build_test_upload.yml#L72-L75
In this use case, you should do `conda create --name ta_python_${{ matrix.python_version }} python=${{ matrix.python_version }} -y

.github/workflows/velox.yml
kgpai: @wenleix the other names in the workflow yml are macos.yml, ubuntu.yml , So this one should be linux-nightly.yml for consistency sake.

.github/workflows/linux-nightly.yml
```diff
@@ -0,0 +1,62 @@
...
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+    steps:
+      - name: Print CPU info
+        run: cat /proc/cpuinfo
+
+      - name: Setup Python environment
+        uses: actions/setup-python@v2
```
wenleix, line 23: Note all the CPython extension are built against Python 3.10, despite you set up Conda with Python 3.7/3.8/3.9:

```
[164/165] Linking CXX shared module csrc/velox/_torcharrow.cpython-310-x86_64-linux-gnu.so
```

Python 3.10 is indeed contained in the PATH and seems to be picked up with higher priority:
```
PATH='\''/opt/conda/bin:/opt/conda/condabin:/opt/conda/envs/ta_python_/bin:/opt/conda/condabin:/__t/Python/3.10.2/x64/bin:/__t/Python/3.10.2/x64:/opt/rh/devtoolset-10/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'\''
```

The Python 3.10 seems to be setup by the Python environment here: 
```
/usr/bin/docker exec  cfcbe2eacbad[6](https://github.com/facebookresearch/torcharrow/runs/5485875682?check_suite_focus=true#step:4:6)31b8a29de181d67682baa7806e69c232673ab8a5a5328fcf2d9 sh -c "cat /etc/*release | grep ^ID"
Successfully setup CPython (3.10.2)
```

wenleix (commented):
Thanks! It looks much simpler with the Python version matrix!

.github/workflows/linux-nightly.yml
ejguan: > The Python 3.10 seems to be setup by the Python environment here. Do we need it (since IIUC, we actually want the Python environment brought by Conda right? )

Agree. This step should be removed.

wenleix (approved):
LGTM. Thanks! Don't forget to revert the Velox submodule change ( you can also revert that after you import it, since it will be just revert a `.txt` file)

PR #253, opened 2022-03-15
[WIP] Add GPT2_BPE_Encoder implementation in torcharrow

Nayef211:
## Description
- Link libtorch against torcharrow
- Add source files for GPT2BPE implementation


csrc/velox/functions/gpt2_bpe_tokenizer.cpp
```diff
@@ -0,0 +1,405 @@
+#include "gpt2_bpe_tokenizer.h"
```
wenleix, line 1: nit: maybe put into `functions/text`?  Simliar to https://github.com/facebookresearch/torcharrow/pull/242/files#diff-17c8133a1b9bc8be7b62d7983c0264777d894e46b0bf6add7c7c10ace70b3d5cR1

setup.py
```diff
@@ -96,7 +96,7 @@ def build_extension(self, ext):
 
         cmake_args = [
             f"-DCMAKE_BUILD_TYPE={cfg}",
-            # f"-DCMAKE_PREFIX_PATH={torch.utils.cmake_prefix_path}",
+            f"-DCMAKE_PREFIX_PATH=/Users/nayef211/opt/miniconda3/envs/torchtext/lib/python3.9/site-packages/torch/share/cmake",
```
wenleix, line 99: you can also use `{torch.utils.cmake_prefix_path}`, similar to TorchAudio:

https://github.com/pytorch/audio/blob/4b47412e91c22d75a1d5f4ea5f71e0ceaa26efd7/tools/setup_helpers/extension.py#L93

csrc/velox/functions/CMakeLists.txt
```diff
@@ -4,12 +4,15 @@
...
 
+
 add_library(
   torcharrow_udfs OBJECT
   functions.h
   string_functions.h
   numeric_functions.h
   register_udf.cpp
+  gpt2_bpe_tokenizer.h
+  gpt2_bpe_tokenizer.cpp
```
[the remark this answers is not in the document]
mthrok: After doing this, you need to bring-in the other source files that `gpt2_bpe_tokenizer` requires, such as `regex.[h|cpp]` and `re2/re2.h`.

csrc/velox/CMakeLists.txt
```diff
@@ -73,16 +73,22 @@ pybind11_add_module(
 add_subdirectory(velox)
 add_subdirectory(functions)
 
+# set(CMAKE_PREFIX_PATH deps/libtorch)
+find_package(Torch REQUIRED)
```
mthrok, line 77: Another possibility that I can think of is that findTorch here is affected by some local settings of TorchArrow. I recommend to move this to the top level CMakeLists or at least very beginning of this file.

.gitmodules
```diff
@@ -5,3 +5,6 @@
 	path = csrc/velox/velox
 	url = https://github.com/facebookincubator/velox.git
 
+[submodule "csrc/velox/re2"]
+	path = csrc/velox/re2
+	url = https://github.com/google/re2.git
```
mthrok, line 10: Velox requires re2. So as long as TorchArrow depends on re2, you do not need to add this submodule. re2 target is already defined. (also I do not see you add something like `add_subdirectory(re2)`

https://github.com/facebookincubator/velox/blob/e4942d0ae32585c48a0faa8f344cdb01bd02ed71/velox/functions/lib/CMakeLists.txt#L14

csrc/velox/functions/gpt2_bpe_tokenizer.cpp
Nayef211: Done!

csrc/velox/functions/functions.h
```diff
@@ -226,6 +227,15 @@ inline void registerTorchArrowFunctions() {
       velox::Array<int64_t>,
       velox::Array<int64_t>>({"bucketize"});
 
+  // TorchText
+  // bpe_encode
+    velox::registerFunction<
+      bpe_tokenize,
+      velox::ArrayWriterT<int64_t>,
+      std::shared_ptr<GPT2BPEEncoder>,
+      velox::Varchar>({"bpe_tokenize"});
```
wenleix, line 236: nit: you can name the struct name as `BpeTokenizeFunction` since it will be aliased to `bpe_tokenize` anyway here.

CMakeLists.txt
```diff
@@ -8,6 +8,6 @@ cmake_minimum_required(VERSION 3.5)
 
 project(torcharrow)
 
-# TODO: Move global build config from csrc/velox/CMakeLists.txt to here
+find_package(Torch REQUIRED)
```
mthrok, line 11: @wenleix Is PyTorch going to be a mandatory dependency of TorchArrow?

wenleix: @mthrok Ideally we want to keep PyTorch as a soft-dep for TorchArrow.

Nayef211:
all changes from this PR are contained in https://github.com/facebookresearch/torcharrow/pull/271

PR #256, opened 2022-03-17
[NOT For Review] Build #253

wenleix:
`python setup.py develop` works now. But since it's not used so some possible linker error may be missed.


csrc/velox/functions/CMakeLists.txt
```diff
@@ -19,4 +21,5 @@ target_link_libraries(
   torcharrow_udfs
   velox_functions_string
   velox_functions_prestosql
+  "${TORCH_LIBRARIES}"
```
wenleix, line 24: `TORCH_LIBRARIES` should be added in the funciton's CMakeList. Otherwise `#include <torch/script.h>` will complain.

csrc/velox/functions/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
```
wenleix, line 8: hope Velox CXX Flags is compatible with Torch CXX flags... (otherwise will be a nightmare)

Velox CXX flags: (https://github.com/facebookincubator/velox/blob/1bdf9fc2a0ba7e9351cc6d092aa5c0e4e0f78708/CMakeLists.txt#L125-L127):

```
-mavx2 -mfma -mavx -mf16c -mbmi2 -march=native
```

PR #259, opened 2022-03-24
Using coalesce for fill_null

bearzx:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #259

Differential Revision: [D35158061](https://our.internmc.facebook.com/intern/diff/D35158061)

PR #260, opened 2022-03-24
[WIP] Using coalesce for fill_null in eager mode

bearzx:
ghstack-source-id: e6c54d1116e27215016e824d6a436efd25b237cd
Pull Request resolved: https://github.com/facebookresearch/torcharrow/pull/259

This PR replaces a python implementation of `fill_null` with velox coalesce UDF.

### DIFF
--- a/torcharrow/test/test_numerical_column.py
+++ b/torcharrow/test/test_numerical_column.py
@@ -423,6 +423,16 @@ class TestNumericalColumn(unittest.TestCase):
         c = c.append([2])
         self.assertEqual(set(c.drop_duplicates()), {None, 2, 17.0})
 
+    def base_test_fill_null_type_promotion_rules(self):
+        c = ta.column([1, 2, 3.0, None], dtype = dt.Float64(nullable=True))
+        d = c.fill_null(4)
+        self.assertEqual(d.dtype, dt.float64)
+        self.assertEqual(list(d), [1.0, 2.0, 3.0, 4.0])
+        c = ta.column([1, 2, 3, None], dtype = dt.Int32(nullable=True))
+        d = c.fill_null(4)
+        self.assertEqual(d.dtype, dt.int32)
+        self.assertEqual(list(d), [1, 2, 3, 4])
+
     def base_test_agg_handling(self):
         import functools
         import operator
--- a/torcharrow/velox_rt/numerical_column_cpu.py
+++ b/torcharrow/velox_rt/numerical_column_cpu.py
@@ -595,23 +595,14 @@ class NumericalColumnCpu(ColumnCpuMixin, NumericalColumn):
     @trace
     @expression
     def fill_null(self, fill_value: Union[dt.ScalarTypes, Dict]):
-        self._prototype_support_warning("fill_null")
-
         if not isinstance(fill_value, Column._scalar_types):
             raise TypeError(f"fill_null with {type(fill_value)} is not supported")
         if not self.is_nullable:
             return self
-        else:
-            col = velox.Column(get_velox_type(self.dtype))
-            for i in range(len(self)):
-                if self._getmask(i):
-                    if isinstance(fill_value, Dict):
-                        raise NotImplementedError()
-                    else:
-                        col.append(fill_value)
-                else:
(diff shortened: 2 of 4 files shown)


torcharrow/velox_rt/numerical_column_cpu.py
```diff
@@ -606,18 +606,8 @@ def fill_null(self, fill_value: Union[dt.ScalarTypes, Dict]):
...
-                if self._getmask(i):
-                    if isinstance(fill_value, Dict):
-                        raise NotImplementedError()
-                    else:
-                        col.append(fill_value)
-                else:
-                    col.append(self._getdata(i))
-            return ColumnCpuMixin._from_velox(self.device, self.dtype, col, True)
+
+        return functional.coalesce(self, fill_value)
```
wenleix, line 610: nit: you need massage around the type nullability. Unfortunately `functional` now pessimistically makes returned column type always nullable, .e. g

```python
return functional.coalesce(self, fill_value)._with_null(False)
```

Perhaps also add a test on returned dtype in test? https://github.com/facebookresearch/torcharrow/blob/0a9eec6149659301cebb007cc409d4b307c382de/torcharrow/test/test_numerical_column.py#L420 ?

wenleix: Also remove the `_prototype_support_warning` on line 603 -- it's now use the native Velox C++ function.

wenleix (dismissed):
LGTM % nits

wenleix:
Should I review this one or https://github.com/facebookresearch/torcharrow/pull/259?

bearzx:
> Should I review this one or #259?

hmm I didn't know why it has a copy ... let's just do this one, I'll drop #259 .

torcharrow/velox_rt/numerical_column_cpu.py
wenleix: Actually , this doesn't work if you have a int32 column and try to fill_null with int (since Python int is interpreted as int64) . (caught by test) . Let's sync offline on this~

wenleix (commented):
LGTM. The velox submodule update seems irrelevant?

bearzx:
> LGTM. The velox submodule update seems irrelevant?

yeah I did a pull and then did a merge, maybe I should start a new clean PR without those?

torcharrow/velox_rt/numerical_column_cpu.py
```diff
@@ -595,23 +595,13 @@ def log(self):
...
-                if self._getmask(i):
-                    if isinstance(fill_value, Dict):
-                        raise NotImplementedError()
-                    else:
-                        col.append(fill_value)
-                else:
-                    col.append(self._getdata(i))
-            return ColumnCpuMixin._from_velox(self.device, self.dtype, col, True)
+
+        fill_value_casted = dt.np_typeof_dtype(self.dtype)(fill_value)
```
wenleix, line 603: add a comment like 
```
# functional doesn't support type promotion yet. Cast to column dtype manually (represented by numpy literal) 
```
? 

Also let's add some basic tests (e.g. `fill_null` test for int32 and float64 column.

wenleix:
> yeah I did a pull and then did a merge, maybe I should start a new clean PR without those?

You should just be able to revert `csrc/velox/velox` by 
```
git checkout HEAD~ csrc/velox/velox
```

torcharrow/test/test_numerical_column.py
```diff
@@ -423,6 +423,12 @@ def base_test_na_handling(self):
         c = c.append([2])
         self.assertEqual(set(c.drop_duplicates()), {None, 2, 17.0})
 
+    def base_test_fill_null_type_promotion_rules(self):
```
wenleix, line 426: You need to add entry in https://github.com/facebookresearch/torcharrow/blob/main/torcharrow/test/test_numerical_column_cpu.py#L12 , otherwise, the test won't run

wenleix (approved):
LGTM. Make sure `csrc/velox/velox` is reverted. You can also amend the imported diff (the diff needs to anyway be landed from Sandcastle)

PR #269, opened 2022-03-26
Initial docs for TorchArrow

wenleix:
Initial skeleton. More docs as future work.


wenleix:
Reference: https://github.com/pytorch/data/pull/193

docs/source/_template/class.rst
```diff
@@ -0,0 +1,9 @@
+.. role:: hidden
+    :class: hidden-section
+.. currentmodule:: {{ module }}
```
wenleix, line 3: @NivekT : Seeing the following warning:
```
/Users/wxie/Code/torcharrow/docs/source/_template/class.rst:3: ERROR: Error in "currentmodule" directive:
maximum 1 argument(s) allowed, 3 supplied.

.. currentmodule:: {{ module }}
WARNING: invalid signature for autoclass ('{{ name }}')
WARNING: don't know which module to import for autodocumenting '{{ name }}' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
```

Wondering if there is anything I am not configured correct? :(

NivekT: That is a template file, and you can use it with `autosummary`. The file where you call needs to have "`.. currentmodule:: ` specified. It doesn't seen like you are using `class.rst` right now?

```rst
.. currentmodule:: torchdata.datapipes.iter

......

.. autosummary::
    :nosignatures:
    :toctree: generated/
    :template: datapipe.rst

    Bz2FileLoader
    Decompressor
    RarArchiveLoader
    TarArchiveLoader
```

wenleix: @NivekT : Right. I see for now I am only using `:template: function.rst`. But see the similar warning. Basically it works fine when used in `autosummary`. But somehow Sphinix try to also parse it alone so such warnings emitted?

docs/source/conf.py
```diff
@@ -0,0 +1,95 @@
...
+html_static_path = ["_static"]
+
+html_css_files = [
+    "css/custom.css",
+]
+
+signature_replacements = {}
+
+def process_signature(app, what, name, obj, options, signature, return_annotation):
+    """Replacing long type annotations in signature with more succinct ones."""
```
wenleix, line 86: @NivekT : noob question about the mechanism of this signature replacement ^_^. Also I assume we can always add that later? (e.g. we don't add that in this initial doc commit)

docs/source/_template/class.rst
NivekT: Yep, I would probably exclude it if you are not using it.

docs/source/conf.py
NivekT: That is entirely optional. I added it because `TorchData` has a need for it.

docs/source/conf.py
```diff
@@ -0,0 +1,95 @@
...
+    if isinstance(signature, str):
+        for old, new in signature_replacements.items():
+            if old in signature:
+                signature = signature.replace(old, new)
+        return signature, return_annotation
+
+
+def setup(app):
+    app.connect("autodoc-process-signature", process_signature)
```
NivekT, line 95: You can take these out if you don't end up using `process_signature` or similar methods. This is optional.

docs/source/conf.py
```diff
@@ -0,0 +1,95 @@
...
+}
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ["_static"]
+
+html_css_files = [
+    "css/custom.css",
+]
```
NivekT, line 81: Also optional, the default `css` capitalizes every letter in any title of your web page. That is incompatible to my need so I added this `custom.css` to overwrite it.

docs/source/conf.py
NivekT: In case you wonder where this comes from:

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#docstring-preprocessing

docs/source/conf.py
wenleix: Got it. Thanks! No wonder why TorchArrow docs renders differently with PyToch doc style :)

NivekT (approved):
Overall, it looks good!

You may want to add a RST template for the  `autosummary` in `dataframe.rst` and `column.rst`. That way the page may render better.

Lastly, you can suppress the warnings by adding this to `conf.py` if you have no plan to render those as standalone pages.
`exclude_patterns = ['_template/class.rst', '_template/function.rst']`

PR #271, opened 2022-03-30
Add UDF for GPT2BPE encoder

Nayef211:
## Description
- Link `libtorch` against torcharrow (original PR https://github.com/facebookresearch/torcharrow/pull/255)
- Add source files for `GPT2BPETokenizer` operator implementation (original PR https://github.com/facebookresearch/torcharrow/pull/253)
- Adding and registering UDF for `GPT2BPETokenizer` operator
- Added Python test to test UDF + operator using dummy assets

## Building
- You need to install PyTorch in your environment
-  `USE_TORCH=1 DEBUG=1 python setup.py develop`

## Testing
`pytest torcharrow/test/test/test_text_ops.py`


csrc/velox/column.cpp
```diff
@@ -623,6 +624,12 @@ velox::variant pyToVariant(const pybind11::handle& obj) {
     return out;
   }
 
+  if (py::isinstance<functions::GPT2BPEEncoder>(obj)) {
+    out = velox::variant::opaque(
+        obj.cast<std::shared_ptr<functions::GPT2BPEEncoder>>());
+    return true;
```
wenleix, line 630: you should return `out` instead of `true`. It expects a variant not a boolean :)

wenleix: or, just simply do 

```c++
return velox::variant::opaque(
        obj.cast<std::shared_ptr<functions::GPT2BPEEncoder>>());
```

Nayef211: This did the trick thank you!

torcharrow/test/test_text_ops.py
```diff
@@ -0,0 +1,124 @@
...
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    return dict(zip(bs, cs))
+
+
+# copied from GPT2BPETokenizer __init__ method
+def init_bpe_encoder():
+    encoder_json_path = get_asset_path("gpt2_bpe_encoder.json")
+    vocab_bpe_path = get_asset_path("gpt2_bpe_vocab.bpe")
```
parmeet, line 60: I wonder if instead of creating these local assets in the TA repo, we could directly rely on downloading it from[ pytorch models repo](https://s3.console.aws.amazon.com/s3/buckets/pytorch?region=us-east-1&prefix=models/text/&showversions=false)? In other words, we could use a temporary directory for testing and use [tochhub API](https://pytorch.org/docs/stable/hub.html#torch.hub.download_url_to_file) to download asset locally for testing.

csrc/velox/CMakeLists.txt
```diff
@@ -69,20 +69,21 @@ pybind11_add_module(
...
-  velox_arrow_bridge
-  torcharrow_udfs
+velox_vector
+velox_memory
+velox_buffer
+velox_core
+velox_functions_prestosql
+velox_function_registry
+velox_arrow_bridge
+torcharrow_udfs
```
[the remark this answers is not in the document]
Nayef211: I think @wenleix suggested that there isn't anything enforced in OSS. But in fbcode they use clang format and black. I will remove these formatting changes from this PR!

csrc/velox/functions/text/regex.h
```diff
@@ -0,0 +1,20 @@
+#include <re2/re2.h>
+#include <re2/stringpiece.h>
+#include <string>
+
+// namespace py = pybind11;
+
+namespace facebook::torcharrow {
```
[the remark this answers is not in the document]
Nayef211: I think this was a mistake on my part. I've been following the convention that torchrec used when adding the bucketize function/UDF. In this case I meant to put all of the operators from torchtext under the `facebook::torcharrow::functions` namespace.

@wenleix lmk if there's a specific convention you want us to adhere to for namespaces.

torcharrow/test/test_text_ops.py
Nayef211: I can give this a shot. The only thing I worry about is that adding a test that requires network access can be problematic when enabling the test within fbcode and we may have to rely on monkey patching to make it work as we do in torchtext. In addition, tests requiring downloads tend to increate the time it takes for all tests to pass in CI. This was one of the reasons we introduced mocking in torchtext to get away from test downloads. Wdyt? 

Also @wenleix lmk if you have an opinion here

parmeet: > The only thing I worry about is that adding a test that requires network access can be problematic when enabling the test within fbcode

Ahh, that's a good point. I didn't think about it when making the proposal. Ya, in that case I think let's go with the asset files. I remember there was discussion on adding a small size vocab for GPT2 (similar to mocking) instead of full model based vocab. Unless @wenleix  see an issue with the file size, I think we are good to go with the current solution :)

Nayef211: Sounds good!

csrc/velox/functions/text/regex.h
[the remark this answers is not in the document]
wenleix: Yeah, `facebook::torcharrow::functions` is the current namespace naming convention (should them because `meta::torcharow::functions` though? :)

PR #276, opened 2022-03-31
Add support for nightly Macos build

kgpai:


wenleix:
For some reason it's using the  Mac default Python? (2.7). Not sure if it relates to conda. 

Note conda is only used to set up a Python environment, and GA can provide Python, so perhaps we can modify this method to work work when `PYTHON_VERSION` is provided: https://github.com/facebookresearch/torcharrow/blob/main/packaging/build_wheel.sh#L45-L50


i.e. 
```
if [[ -n "$PYTHON_VERSION" ]]; then
     eval "$(conda shell.bash hook)"
     ...
fi
```

So you don't need conda when building wheel?

.github/workflows/linux-nightly.yml
```diff
@@ -54,3 +55,49 @@ jobs:
           for pkg in fixed_dist/torcharrow*.whl; do
               ~/.local/bin/aws s3 cp "$pkg" "$S3_PATH" --acl public-read
           done
+
+  macos-container:
+    runs-on: macos-11
```
wenleix, line 60: shall we just use macos-latest?

kgpai: i tried that too and it defaults to macos-11 , also atm macos 12 breaks velox (we are fixing it) , so I dint want to set it to latest just yet.

.github/workflows/linux-nightly.yml
```diff
@@ -54,3 +55,47 @@ jobs:
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Check out source repository
+        uses: actions/checkout@v2
+        with:
+            submodules: true
+
+      - name: Build the wheel
+        run: |
+          ./csrc/velox/velox/scripts/setup-macos.sh
```
wenleix, line 81: unrelated to this diff; But would be great for Velox repo to provide and maintain the "setup-{macos/ubuntu}-minimal.sh"

wenleix (approved):
LGTM

.github/workflows/linux-nightly.yml
```diff
@@ -54,3 +55,47 @@ jobs:
...
+          pip install wheel
+          ./packaging/build_wheel.sh
+          pip install delocate
+          delocate-wheel dist/*.whl -w fixed_dist
```
ejguan, line 85: Technically, we should install `whl` and run tests before uploading to nightly.

packaging/build_wheel.sh
```diff
@@ -43,10 +43,13 @@ setup_macos() {
...
-  eval "$(conda shell.bash hook)"
-  conda env remove -n "env$PYTHON_VERSION" || true
-  conda create -yn "env$PYTHON_VERSION" python="$PYTHON_VERSION"
-  conda activate "env$PYTHON_VERSION"
+  if [[ -n "$PYTHON_VERSION" ]]; then
+    eval "$(conda shell.bash hook)"
+    conda env remove -n "env$PYTHON_VERSION" || true
+    conda create -yn "env$PYTHON_VERSION" python="$PYTHON_VERSION"
+    conda activate "env$PYTHON_VERSION"
+ fi
```
ejguan, line 51: Noob question: Do we need to install `conda` before this step?

wenleix: nit: the indentation of this line look incorrect

wenleix: @ejguan If you want to use Conda to set up Python environment, then yes. So that's why I say "if PYTHON_VERSION is unset, don't use conda.."

.github/workflows/linux-nightly.yml
wenleix: @ejguan : Right. But the GA Mac machine is old cpu and doesn't have AVX2, it's guaranteed to have "illegal instructions". 

Another reason why make Velox to support non-AVX2 is so critical ^_^. cc @Yuhta, @pedroerp , @damianr99

.github/workflows/linux-nightly.yml
kgpai: Yes makes sense . Will do that as a part of kr3.

PR #280, opened 2022-04-03
Install latest PyTorch for testing

wenleix:
Otherwise, PyTorch install step is failing due to
https://discuss.pytorch.org/t/tls-cert-for-download-pytorch-org-expired/148169

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Install dependencies for test
         run: |
           pip3 install pytest  --user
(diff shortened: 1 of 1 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       - name: Install dependencies for test
         run: |
           pip3 install pytest  --user
-          pip3 install torch==1.10.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html --user
+          pip3 install torch --extra-index-url https://download.pytorch.org/whl/cpu/torch_stable.html --user
```
wenleix, line 75: based on the command in https://pytorch.org/get-started/locally/

PR #282, opened 2022-04-05
Build on Mac with development target 10.15

wenleix:
Other PyTorch domain libraries still publish wheels that supports 10.X (e.g. 10.9 or 10.15)


wenleix:
Landed via d45d86231cb80963762dffe83d5454edeedb97bb

PR #303, opened 2022-04-28
Conda build recipe for macosx

bearzx:
Initial recipe to enable conda build for macosx, after a lot of trial and error.

The main body of the recipe is borrowed from torch data, torch vision and torch text, with tweaks to make it work with torcharrow (the main challenge here is to compile velox). Few things that are worth mentioning:

MACOSX_DEPLOYMENT_TARGET needs to be on 10.15 so that certain c++ features used in velox can be recognized.
python needs to be in all build, host, and run requirements, otherwise you might run into issues that different pythons are used in different phases.
boost is required in both build and run, I ran into some weird errors without having both.
I think the scripts/build_mac_dep.sh step is still needed as some packages are installed with source files (e.g. folly). I don't know if it's possible to incorporate everything into conda requirements. This is something we can explore further.
To build the package:

cd torcharrow
PYTHON_VERSION=3.7 ./packaging/build_conda.sh
To install

conda install -c file:///Users/your_user_name/torcharrow/conda-bld -c pytorch-nightly torcharrow
The -c pytorch-nightly is critical because it'll help to find all the dependencies.

Couple of things left TODO:

unittest: by default conda will copy torcharrow/test into the package, but it doesn't run them. To run, a run_test.sh needs to be created. I actually did that, but I ran into some segmentation fault errors when running certain tests. My guess is it has something to do with compiling velox, but seems like it'll need further investigation.
linux and windows build are not included here. I put the linux requirements in the recipe, but I didn't test those. This will be done in followup efforts.
but even just for macosx build, this also needs to be enabled on the github workflow.


packaging/torcharrow/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - libdouble-conversion-dev
+    - libgoogle-glog-dev
+    - libgflags-dev
+    - libevent-dev
+    - libre2-dev
+    {% endif %}
+  host:
+    - python
+    - setuptools
+    - cpuonly
```
ejguan, line 42: You don't need `cpuonly` if `torch` is not a dependency.

packaging/torcharrow/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+  script_env:
+    - BUILD_VERSION
+
+test:
+  imports:
+    - torcharrow
+  source_files:
+    - torcharrow/test
+  requires:
+    - cpuonly
```
ejguan, line 69: Ditto

packaging/torcharrow/build.sh
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set -ex
+
+git submodule update --init --recursive
+scripts/build_mac_dep.sh ranges_v3 fmt double_conversion folly re2
```
ejguan, line 11: Is this intentional to only build on mac?

ejguan (commented):
I have a few comments below. In general, looks good.

packaging/torcharrow/build.sh
bearzx: actually on my local copy I often comment out this line when running the build, because I already have the dependencies installed and there is a `prompt` command in `build_mac_dep.sh` that doesn't run very well with the conda build process.

ejguan: I thought conda-build would create a new isolated environment and you have to install all deps within this env.

bearzx: yep, so this script should work without running part with `prompt` if it's ran on a fresh build environment. I'll update the PR with this line removed and later add it to the github action workflow.

ejguan: Cool. LMK when you have gha created. Then, we can see which part is missing.

PR #315, opened 2022-05-05
Do not use arch=native flag for velox.

kgpai:
GCC can be very aggressive and use avx512 in that case.


csrc/velox/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED True)
 # velox/common/base/SimdUtil.h requires AVX
 # TODO: only need to retain -mavx2?
 set(CMAKE_CXX_FLAGS
-    "${CMAKE_CXX_FLAGS} -mavx2 -mfma -mavx -mf16c -mbmi2 -march=native")
+    "${CMAKE_CXX_FLAGS} -mavx2 -mfma -mavx -mf16c -mbmi2 -mbmi")
```
wenleix, line 28: we need both bmi2 and bmi?

wenleix (approved):
LGTM. Can you double check on the velox submodule update ? -- it says nothing to compare, and it seems to update all the files (625): https://github.com/facebookincubator/velox/compare/95c89c4d88611c6c68899d397015d70cc8810cd4...bb62f5511efa1d34668424e483448b94b1904ce8

csrc/velox/CMakeLists.txt
kgpai: Thats right.

.github/workflows/macos.yml
```diff
@@ -28,7 +28,9 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Update submodules
-        run: git submodule update --init --recursive
+        run: |
+          git submodule sync --recursive
```
wenleix, line 32: I think `actions/checkout` automtaically checkout the submodule recursively: 

```
  /usr/bin/git submodule foreach --recursive git config --local --name-only --get-regexp 'http\.https\:\/\/github\.com\/\.extraheader' && git config --local --unset-all 'http.https://github.com/.extraheader' || :
```

wenleix: I guess a better approach would be 

```
      - name: Check out source repository
        uses: actions/checkout@v2
        with:
            submodules: recursive
```

similar to https://github.com/facebookresearch/torcharrow/blob/6737648fdd081f6b5a46ce0d041dca3e80ea50f1/.github/workflows/nightly-wheel.yml#L27-L30

kgpai: yes we used this for nightly build..

wenleix:
cc @ejguan , @NivekT : This should fix the illegal instruction issues on Intel(R) Xeon(R) CPU E5-2673 v3/v4 (basically CPU with AVX2 but not AVX512)

wenleix:
cc @YLGH

PR #319, opened 2022-05-07
Conda nightly build to Github Actions

bearzx:
In this PR I'm added a new Github workflow to enable nightly build for conda. But right now for developing and testing, I left the trigger as new PR comes in, so it's not really "nightly build" yet.

Multiple points to note:
* Since Github Action machines mac don't support AVX instructions, I left the test blank for mac builds, and only enabled it for linux build.
* For some reason, glog needs to be on 0.4.0 version (latest 0..5.0) to make the testing part work (even it's just a simple import), otherwise some symbol not found error will arise. I think it has something to do with glog itself, so I'm leaving the workaround as is for now.
* I actually tried to run the unit tests with linux build, and it mostly worked except for two that need to check package names. I think when running tests from conda the top level folder is just `torcharrow`, so the failing tests will have messages like "torcharrow.test" != "test". I'll leave this to further investigation and depending on whether we really want to run unittests in conda build workflows, we may end up not bothering with it.

This PR finishes the building part, but I still need to upload the built artifact to conda, which I'll leave as a todo.

### DIFF
--- a/.github/workflows/nightly-conda.yml
+++ b/.github/workflows/nightly-conda.yml
@@ -1,0 +1,113 @@
+name: TorchArrow Nightly Build and Test for Conda
+on:
+#  schedule:
+#    - cron: "0 0 * * *"
+  pull_request:
+    branches:
+      - main
+      # For PR created by ghstack
+      - gh/*/*/base
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  linux-container:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+    steps:
+      - name: Print CPU info
+        run: |          
+          cat /proc/cpuinfo          
+
+      - name: Create Conda Env
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+          activate-environment: conda_build_env
+
+      - name: Check out source repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      # Based on https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
+      - name: Prepare ccache timestamp
+        id: ccache_cache_timestamp
+        run: |
+          echo "::set-output name=timestamp::$(/bin/date -u "+%Y%m%d-%H:%M:%S")"
+        shell: bash
+
+      - name: Load ccache files
+        uses: actions/cache@v2
+        with:
+          path: .ccache
+          key: ubuntu-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+          restore-keys: |
+              ubuntu-ccache-      
+
+      - name: Install dependencies with APT
+        run: |
+          sudo apt-get update
+          sudo apt install -y g++ cmake ccache ninja-build checkinstall git \
+          libssl-dev libboost-all-dev libdouble-conversion-dev libgoogle-glog-dev \
+          libgflags-dev libevent-dev libre2-dev
+
+      # Based on https://github.com/facebookincubator/velox/blob/99429407c3d524e07b32b8b19a03aa7382f819cf/.circleci/config.yml#L114-L116
+      - name: Configure ccache
+        run: |
+          echo "$GITHUB_WORKSPACE"
+          CCACHE_DIR=$GITHUB_WORKSPACE/.ccache_root ccache -sz -M 1G
+          CCACHE_DIR=$GITHUB_WORKSPACE/.ccache ccache -sz -M 1G      
+
+      - name: Build and install folly and fmt
+        # sudo doesn't preserve environment vairable; set it after sudo: https://stackoverflow.com/questions/8633461/how-to-keep-environment-variables-when-using-sudo/33183620#33183620
+        run: |
+          sudo CMAKE_C_COMPILER_LAUNCHER=ccache CMAKE_CXX_COMPILER_LAUNCHER=ccache CCACHE_DIR=$GITHUB_WORKSPACE/.ccache_root scripts/setup-ubuntu.sh
+
+      - name: Conda Build
+        shell: bash -l {0}
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly-conda.yml
```diff
@@ -0,0 +1,120 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+  workflow_call:
+    secrets:
+      AWS_PYTORCH_UPLOADER_ACCESS_KEY_ID:
+        required: true
+      AWS_PYTORCH_UPLOADER_SECRET_ACCESS_KEY:
+        required: true
```
ejguan, line 19: You don't need this part. `workflow_call` is used to invoke this workflow script from another workflow.

packaging/torcharrow/meta.yaml
```diff
@@ -11,49 +11,39 @@ requirements:
...
+    - glog 0.4.0
     {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
 
 build:
   string: py{{py}}
   script_env:
     - BUILD_VERSION
 
+#GitHub Actions usually provide Mac without AVX2 support, for now just skip the test until Velox can be run without AVX2
+#pytest --no-header -v torcharrow/test
```
ejguan, line 37: I think Velox team has solved this issue.

bearzx: nope the issue was on some machines with AVX2 support velox still encounters issues sometimes, but if the machine doesn't support AVX2 in the first place (which is the case for Github Action mac machines), it's guaranteed to fail.

bearzx: you mean the ones with AVX2 support but still fails? I haven't met those on Github Actions (since they don't support AVX2 anyways). Here's an early test that I did have the "illegal instruction" errors, it happened when doing the import torcharrow after building: https://github.com/facebookresearch/torcharrow/runs/6297806435?check_suite_focus=true . Please note that the workflow file is different from the one in this PR, but the main structure for macos-container should be almost the same.

wenleix: Yeah, https://github.com/facebookresearch/torcharrow/pull/315 and https://github.com/facebookincubator/velox/pull/1549 solves the issue that Velox requires AVX512. But still needs AVX2 :)

kgpai:
@wenleix  we explicitly depend on glog v 0.4.0 .  I havent upgraded to 0.5.0 to see if they are compatible, but based on error it probably isnt.

packaging/torcharrow/meta.yaml
kgpai: We can actually support cpus without avx now (thats how we will support m1 etc). Our build scripts havent caught up with it, but I will make those changes shortly.

bearzx:
> @wenleix we explicitly depend on glog v 0.4.0 . I havent upgraded to 0.5.0 to see if they are compatible, but based on error it probably isnt.

Yeah building itself seems to be fine, but when testing it'll need glog to be on 0.4.0 as 0.5.0 removed some non-public symbols, which caused errors. Here's a discussion for this issue that I found: https://trac.macports.org/ticket/62921

packaging/torcharrow/meta.yaml
bearzx: nice, I'll enable the mac tests when it's ready.

.github/workflows/nightly-conda.yml
```diff
@@ -0,0 +1,113 @@
...
+          submodules: recursive
+
+      # Based on https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
+      - name: Prepare ccache timestamp
+        id: ccache_cache_timestamp
+        run: |
+          echo "::set-output name=timestamp::$(/bin/date -u "+%Y%m%d-%H:%M:%S")"
+        shell: bash
+
+      - name: Load ccache files
```
wenleix, line 47: @ejguan in general, do we prefer to use ccache for packaging, or prefer to "clean build" from scratch? ~

packaging/torcharrow/meta.yaml
```diff
@@ -11,49 +11,39 @@ requirements:
...
     - cffi
+    - glog 0.4.0
     {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
 
 build:
   string: py{{py}}
   script_env:
     - BUILD_VERSION
 
+#GitHub Actions usually provide Mac without AVX2 support, for now just skip the test until Velox can be run without AVX2
```
wenleix, line 36: noob question: is this file read by the command like `conda build`?

wenleix (approved):
LGTM. But @ejguan may have more specific knowedge about Conda build ^_^

packaging/torcharrow/meta.yaml
```diff
@@ -11,49 +11,39 @@ requirements:
...
-    - boost
-    - libboost
     - typing_inspect
     - pandas
     - typing
     - tabulate
     - pyarrow
     - arrow
     - cffi
+    - glog 0.4.0
```
ejguan, line 28: Log from your workflow:
```
Adding .* to spec 'glog 0.4.0' to ensure satisfiability.  Please consider putting {{ var_name }}.* or some relational operator (>/</>=/<=) on this spec in meta.yaml, or if req is also a build req, using {{ pin_compatible() }} jinja2 function instead.  See https://conda.io/docs/user-guide/tasks/build-packages/variants.html#pinning-at-the-variant-level
WARNING:conda_build.utils:Adding .* to spec 'glog 0.4.0' to ensure satisfiability.  Please consider putting {{ var_name }}.* or some relational operator (>/</>=/<=) on this spec in meta.yaml, or if req is also a build req, using {{ pin_compatible() }} jinja2 function instead.  See https://conda.io/docs/user-guide/tasks/build-packages/variants.html#pinning-at-the-variant-level
```
I think it might be better to use do `glog==0.4.0`

packaging/torcharrow/meta.yaml
ejguan: Yes. This is the file to specify all configuration of build/test/run for conda package.

ejguan (approved):
I am good with conda package related files with one minor comment.

packaging/torcharrow/meta.yaml
bearzx: nice catch! Will update

PR #328, opened 2022-05-11
Use the Dense feature conversion in Criteo test

wenleix:
This calls into the native C++ implementation

https://github.com/facebookresearch/torcharrow/blob/c391ad30d2d891d05ab165035a40a4b9ef88df3f/csrc/velox/tensor_conversion.cpp#L15-L33


bearzx (approved):
looks good to me

PR #336, opened 2022-05-19
Tiny tweak on the feature generation threshold

bearzx:
Previously the dense feature looks more sparse than the sparse feature. I replaced the two to avoid confusion.


bearzx:
actually the sparse here doesn't mean the feature vector appears to be sparse, it means the cardinality of the feature is high (i.e. the range the values is wide)

ifunanyaScript:
Exactly!

PR #340, opened 2022-05-20
Fix conda install crash

bearzx:
~~In this PR I removed the version constraint on pytorch. With it, we'll run into some version mismatches when doing `conda install`. Since we don't strongly depend on the nightly build of pytorch itself, this should be a safe change.~~

Actually after discussing with @ejguan, I realized I can simply remove the pytorch nightly dep in the `host` section, and only leave it in the `run` section, as we don't really need pytorch when building. I had it tested and it worked well.

### DIFF
--- a/packaging/torcharrow/meta.yaml
+++ b/packaging/torcharrow/meta.yaml
@@ -12,7 +12,6 @@ requirements:
   host:
     - python
     - setuptools
-    # {{ environ.get('CONDA_PYTORCH_BUILD_CONSTRAINT') }}
     - pytorch
   run:
     - python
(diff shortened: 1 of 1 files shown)


packaging/torcharrow/meta.yaml
```diff
@@ -12,7 +12,7 @@ requirements:
   host:
     - python
     - setuptools
-    {{ environ.get('CONDA_PYTORCH_BUILD_CONSTRAINT') }}
+    - pytorch
```
ejguan, line 15: If not, why not just remove `pytorch` from the dependency?

bearzx: yeah good point, I copied it from the other recipes, but now that I think of it, building probably doesn't need pytorch. Let me test it.

ejguan (approved):

wenleix (approved):
LGTM.

PR #343, opened 2022-05-23
Add docs for column

wenleix:


docs/source/column.rst
```diff
@@ -41,11 +60,58 @@ NumericalColumn class reference
...
+
+StringColumn class reference
+-----------------------------------
+.. class:: StringColumn()
 
+.. autosummary::
+    :toctree: generated
+    :nosignatures:
+
+    torcharrow.istring_column.StringMethods.length
```
wenleix, line 80: tried to use `StringColumn.str.length` but seems Sphinx doesn't work. @NivekT should I just use `istring_column.StringMethods.length` and some how do string substitution in doc?

NivekT: Maybe try this?
https://stackoverflow.com/questions/20569011/python-sphinx-autosummary-automated-listing-of-member-functions

NivekT: It should allow you to show methods under class. Let me know if that helps

wenleix: @NivekT I mean, the problem is `StringColumn.str` is a member in `StringColumn`~

ejguan (approved):

docs/source/column.rst
ejguan: How about using alias in TA?
`torcharrow.StringColumn.str.length = torcharrow.istring_column.StringMethods.length`

NivekT: Yea in that case, the substitution that you linked may be the most straightforward way.

torcharrow/istring_column.py
```diff
@@ -203,6 +206,7 @@ def count(self, pat: str):
 
     @abstractmethod
     def find(self, sub):
+        """Return lowest indexes in each strings in the Column."""
```
bearzx, line 209: indexes -> indices?

docs/source/column.rst
bearzx: here the substitution only changes the signature right? We can't make the change to function name

NivekT: How does it currently look and how would you like it to be?

You might get a shorter name if you change `.. currentmodule::` to something else.

bearzx: Currently it looks like this:
Screen Shot 2022-06-16 at 10 37 34 AM

We want to shorten the `torcharrow.istring_column.StringMethods` part to `StringColumn.str.length`, is that right? @wenleix . Since this is the function name not signature, the process-signature hook won't help here.

NivekT: Can you create an alias in python using import statement then show the aliases in the doc?

For example, in `TorchData` we do the following:
```python
from torch.utils.data.datapipes.iter.callable import (
    CollatorIterDataPipe as Collator,
    MapperIterDataPipe as Mapper,
)
```
It shows up as this in our doc:

Screen Shot 2022-06-16 at 3 09 24 PM

bearzx: > ```python
> from torch.utils.data.datapipes.iter.callable import
> ```

where is this code located?

NivekT: https://github.com/pytorch/pytorch/blob/master/torch/utils/data/datapipes/iter/__init__.py

NivekT: Also this part within `TorchData`'s`conf.py` may be worth looking into:

```python
def setup(app):

    # Overwrite class name to allow aliasing in documentation generation
    import torchdata.datapipes.iter as iter
    import torchdata.datapipes.map as map

    for mod in (iter, map):
        for name, obj in mod.__dict__.items():
            if isinstance(obj, type):
                obj.__name__ = name

    app.connect("autodoc-process-signature", process_signature)
```

PR #347, opened 2022-05-24
Create SSE build for torcharrow

kgpai:
This creates a nightly build with SSE for torcharrow on macs.


kgpai:
@wenleix  Do we need a separate SSE build for macs ? I went via this route since I dont think cpu arch flags are captured in the wheels and thus was afraid the binaries would clobber. Let me know if you think it should be a separate build.

.github/workflows/nightly-wheel.yml
```diff
@@ -1,7 +1,7 @@
 name: TorchArrow Nightly Build and Test
 on:
-  schedule:
-    - cron: "0 0 * * *"
+  push:
```
kgpai, line 3: This is just for test purposes.

.github/workflows/nightly-wheel.yml
```diff
@@ -77,7 +78,7 @@ jobs:
 
       - name: Build the wheel
         run: |
-          MACOSX_DEPLOYMENT_TARGET=10.15 ./csrc/velox/velox/scripts/setup-macos.sh
+          MACOSX_DEPLOYMENT_TARGET=10.15 CPU_TARGET="sse" ./csrc/velox/velox/scripts/setup-macos.sh
```
wenleix, line 81: do we also want to set CPU target when build wheel? Also what about Ubuntu wheel?

wenleix (commented):
Thanks. This is great!

Can we also run the mac/linux CI with `CPU_TARGET="sse"` ?-- So we can enable the Mac CI (currently disabled): 
https://github.com/facebookresearch/torcharrow/blob/929c57032093e441b0df719114ab968ace755692/.github/workflows/macos.yml#L73-L77

wenleix:
@kgpai 
> Do we need a separate SSE build for macs ? I went via this route since I dont think cpu arch flags are captured in the wheels and thus was afraid the binaries would clobber. Let me know if you think it should be a separate build.

Yeah I would say let's create and test SSE binary (which is most portable) for now. And we will support runtime dispatch based on arch in the future.

.github/workflows/nightly-wheel.yml
kgpai: Yes, I am adding cpu target when building wheel. I can make ubuntu wheel also see if required.

kgpai: Fyi ubuntu sse wheel will require creating a new docker image so will create separate pr for that.

bearzx:
also kicked off a test on the conda build here: https://github.com/facebookresearch/torcharrow/actions/runs/2380658257

kgpai:
@bearzx FYI, This shouldnt solve the dylib problem, only enables sse builds..

bearzx:
@kgpai btw do we also want to update the scripts in /csrc/velox/velox/scripts ?

kgpai:
@bearzx They are already updated , unless you have something else in mind.

bearzx:
> also kicked off a test on the conda build here: https://github.com/facebookresearch/torcharrow/actions/runs/2380658257

seems like the linux build will fail on a undefined symbol error:

```
undefined symbol: _ZN5folly3f146detail12F14LinkCheckILNS1_17F14IntrinsicsModeE2EE5checkEv
```

probably because `CPU_TARGET="sse"` needs to be added in the comda build workflow. I can update it in another PR after this one lands.

bearzx (approved):

kgpai:
@wenleix @bearzx all tests passed and it published, so reverting nightly runs to scheduled..

PR #351, opened 2022-05-25
Add sse build to conda nightly build

bearzx:
As titled. Since now we have it available with velox, this PR just completes the Conda counterpart. 

We only need to simply pass a CPU_TARGET=sse env var before running the setup.py, but I spent a little bit of time understanding how the env vars are passed from conda meta.yaml recipe to the building script. It seems like from `conda build` to `build.sh` the variables in meta.yaml will be lost. Based on the [conda document](https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html) the best practice is probably to put the env vars in build.sh as well.

This PR also refactored a couple of other related env vars into meta.yaml, and enabled the import test for macos builds.


ejguan (approved):
PR LGTM.

Got a question regarding
> It seems like from conda build to build.sh the variables in meta.yaml will be lost. Based on the conda document the best practice is probably to put the env vars in build.sh as well.

Curious about this statement here. For TorchData, the env `BUILD_S3` is specified by workflow env. And, it seems this variable is passed into the packaging script and `conda build`
https://github.com/pytorch/data/blob/3d967c63450b561355bdce719a0db1950ac1f9eb/.github/workflows/_build_test_upload.yml#L254-L258

bearzx:
> PR LGTM.
> 
> Got a question regarding
> 
> > It seems like from conda build to build.sh the variables in meta.yaml will be lost. Based on the conda document the best practice is probably to put the env vars in build.sh as well.
> 
> Curious about this statement here. For TorchData, the env `BUILD_S3` is specified by workflow env. And, it seems this variable is passed into the packaging script and `conda build` https://github.com/pytorch/data/blob/3d967c63450b561355bdce719a0db1950ac1f9eb/.github/workflows/_build_test_upload.yml#L254-L258

From my experience it can be passed to `build_conda.sh`, but will be lost in `build.sh` which is called by `conda build`. Let me run a couple of more tests to confirm.

bearzx:
> > PR LGTM.
> > Got a question regarding
> > > It seems like from conda build to build.sh the variables in meta.yaml will be lost. Based on the conda document the best practice is probably to put the env vars in build.sh as well.
> > 
> > 
> > Curious about this statement here. For TorchData, the env `BUILD_S3` is specified by workflow env. And, it seems this variable is passed into the packaging script and `conda build` https://github.com/pytorch/data/blob/3d967c63450b561355bdce719a0db1950ac1f9eb/.github/workflows/_build_test_upload.yml#L254-L258
> 
> From my experience it can be passed to `build_conda.sh`, but will be lost in `build.sh` which is called by `conda build`. Let me run a couple of more tests to confirm.

Ok after some tests I think I figured it out: Env vars need to be specified in `meta.yaml -> build -> script_env` to be preserved during build. But it seems like it has some predefined ones already, like `MACOSX_DEPLOYMENT_TARGET`. One thing that confused me was I printed out the variables, but some have values while some don't.

.github/workflows/nightly-conda.yml
```diff
@@ -69,14 +69,20 @@ jobs:
           CCACHE_DIR=$GITHUB_WORKSPACE/.ccache ccache -sz -M 1G
 
       - name: Build and install folly and fmt
+        env:
+          CMAKE_C_COMPILER_LAUNCHER: ccache
```
wenleix, line 73: nice :)

wenleix (approved):
Thanks!

wenleix:
> Based on the conda document the best practice is probably to put the env vars in build.sh as well.

Would also be great to put the conda document link?

PR #355, opened 2022-06-02
Fix velox dylib loading paths for conda build on macos

bearzx:
In this PR I'm attempting to introduce a temporary workaround to fix the velox dylib issue, via a post-build hack. This is to make `conda install torcharrow` work out of the box. 

Currently, you will encounter an error of dylibs not found when velox is dlopen-ing them. This is because the location of those dylibs were determined at compile time. For example: `/usr/local/opt/gflags/lib/libgflags.2.2.dylib` was installed to a default location by `brew install`.

We can specify the dylib dependencies in the conda recipe, but they will be installed to the conda environment locations, rather than the system ones (`/user/local/...`). As a result, the velox dylib (`_torcharrow.*.so`) won't be able to find those libraries.

On macos there is a tool called `install_name_tool` that can change the dylib paths, my plan is to modify those dylib paths from absolute ones to `@loader_path/...` so that the velox .so can find the dylibs installed by conda.

### DIFF
--- a/.github/workflows/nightly-conda.yml
+++ b/.github/workflows/nightly-conda.yml
@@ -136,6 +136,29 @@ jobs:
           scripts/build_mac_dep.sh ranges_v3 fmt double_conversion folly re2
           ./packaging/build_conda.sh
 
+      - name: Fix Velox Dylib Paths
+        shell: bash -l {0}
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+        run: |
+          cd conda-bld/osx-64
+          pkg_name=`ls ./torcharrow-* | sed -n -e 's/.*\(torcharrow.*\).tar.bz2/\1/p'`
+          mkdir ./${pkg_name}
+          tar -xf ./${pkg_name}.tar.bz2 -C ./${pkg_name}
+          rm ./${pkg_name}.tar.bz2
+          cd ./${pkg_name}
+
+          source ${GITHUB_WORKSPACE}/packaging/fix_conda_dylib_paths.sh
+          conda_lib_folder=lib/python${PYTHON_VERSION}/site-packages/torcharrow
+          so_name=`ls ${conda_lib_folder}/_torcharrow.* | sed -n -e 's/.*\(_torcharrow.*\.so\)/\1/p'`
+          fix_velox_dylib_paths ${conda_lib_folder}/${so_name}
+          otool -L ${conda_lib_folder}/${so_name}
+
+          tar -cjf ../${pkg_name}.tar.bz2 *
+          cd ..
+          rm -rf ./${pkg_name}
+          cd ../..
+
       - name: Conda Upload
         shell: bash -l {0}
         env:
--- a/packaging/fix_conda_dylib_paths.sh
+++ b/packaging/fix_conda_dylib_paths.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+function fix_velox_dylib_paths() {
+  velox_so=$1
+
+  # other libs
+  libs="libgflags libglog libz libssl libcrypto libbz2 liblz4 libzstd libsodium"
+
+  for libx in ${libs}
+  do
+    liby=$(otool -L "${velox_so}" | sed -n -e "s/\\(.*${libx}\..*dylib\\).*/\\1/p" | tr -d '[:blank:]')
+
+    install_name_tool -change "${liby}" "@loader_path/../../../${libx}.dylib" "${velox_so}"
+  done
+
+  libx=libevent
+  # libevent
+  liby=$(otool -L "${velox_so}" | sed -n -e "s/\\(.*${libx}-.*dylib\\).*/\\1/p" | tr -d '[:blank:]')
+  install_name_tool -change "${liby}" "@loader_path/../../../${libx}.dylib" "${velox_so}"
+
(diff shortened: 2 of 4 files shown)


packaging/torcharrow/meta.yaml
```diff
@@ -24,16 +24,15 @@ requirements:
...
+    - pytorch
+    # {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
 
 build:
   string: py{{py}}
   script_env:
     - BUILD_VERSION
     - CPU_TARGET
 
-#GitHub Actions usually provide Mac without AVX2 support, for now just skip the test until Velox can be run without AVX2
```
wenleix, line 35: so the test is already enabled now, right?

packaging/torcharrow/meta.yaml
```diff
@@ -24,16 +24,15 @@ requirements:
     - cffi
     - glog==0.4.0
     - libsodium
-    {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
+    - pytorch
```
wenleix, line 27: curious: why do we need to build with pytorch? (since we don't build text UDFs by default)

bearzx: this is in the run section, which specifies the dependencies needed to use torcharrow. I think torcharrow still imports pytorch right?

bearzx:
Ok after a million times of trial-and-error, I finally found a way to adjust the recipe to 
1. Run the conda build as usual.
2. Use `install_name_tool` to change the dylib paths for _torcharrow.so post build.
3. Specify the right versions of runtime dependencies that velox is happy with (notably glog==0.6.0 and arrow-cpp==8.0.0).

The most tricky part is perhaps in step 2, because conda build doesn't provide a good entry point to do a final change before it packages everything into a .tar.bz2 file. Adding stuff after `python setup.py` in `build.sh` sort of worked, but there are some extra steps to specify the compatible dylibs versions after, so it will still end up creating broken build (I still don't know all the details here, just making the observations based on my experiments).

Since the conda package is just a .tar.bz2 file, the approach I ended up taking was to unpack it, install_name_tool, pack it again using the `tar` command, before uploading happens. One more detail here: we'll need to run `tar` within the target folder so that the file list doesn't have a `./` prefix, this is critical because `anaconda upload` [specifically hard-coded](https://github.com/Anaconda-Platform/anaconda-client/blob/be1e14936a8e947da94d026c990715f0596d7043/binstar_client/utils/detect.py#L71) this to read the metadata. It sounds and works pretty hacky, but might be fine as a temp fix before we have static link option available on velox.

bearzx:
Test CI run: https://github.com/facebookresearch/torcharrow/runs/6764330823

packaging/torcharrow/meta.yaml
```diff
@@ -20,20 +20,22 @@ requirements:
     - typing
     - tabulate
     - pyarrow
-    - arrow
```
bearzx, line 23: btw `arrow` on conda is a completely different lib ...

wenleix: right , or `pyarrow` i guess? https://arrow.apache.org/docs/python/install.html

packaging/torcharrow/meta.yaml
```diff
@@ -20,20 +20,22 @@ requirements:
...
+    - arrow-cpp==8.0.0
     - cffi
-    - glog==0.4.0
+    - glog==0.6.0
```
wenleix, line 25: anything changed on Velox side?

packaging/fix_conda_dylib_paths.sh
```diff
@@ -0,0 +1,33 @@
+function fix_velox_dylib_paths() {
+  velox_so=$1
```
wenleix, line 2: nit: can we check there is exactly one parameter ?

.github/workflows/nightly-conda.yml
```diff
@@ -136,6 +136,29 @@ jobs:
...
+          PYTHON_VERSION: ${{ matrix.python-version }}
+        run: |
+          cd conda-bld/osx-64
+          pkg_name=`ls ./torcharrow-* | sed -n -e 's/.*\(torcharrow.*\).tar.bz2/\1/p'`
+          mkdir ./${pkg_name}
+          tar -xf ./${pkg_name}.tar.bz2 -C ./${pkg_name}
+          rm ./${pkg_name}.tar.bz2
+          cd ./${pkg_name}
+
+          source ../../../packaging/fix_conda_dylib_paths.sh
```
wenleix, line 151: can we just use `${GITHUB_WORKSPACE}/packging/...`?

wenleix (approved):
LGTM.

packaging/torcharrow/meta.yaml
bearzx: I think velox always uses the glog installed from brew, which was recently updated from 0.5.0 to 0.6.0. Building with glog0.6.0 and running with glog0.4.0 will not work. However, the conda default channel hasn't updated to 0.6.0 yet, it only exists on conda-forge, hence this change.

bearzx: This is also why we need to specify `-c conda -forge` when installing.

packaging/fix_conda_dylib_paths.sh
bearzx: can you elaborate?

packaging/torcharrow/meta.yaml
bearzx: and `arrow-cpp`

packaging/torcharrow/meta.yaml
ejguan: You might need to add command `pytest --no-header -v torcharrow/test` back the test

PR #361, opened 2022-06-03
Improve user experience for PadSequence collation

wenleix:
Velox functional today always mark dtype as nullable,
make the collation support nullable dtype.

### DIFF
--- a/torcharrow/velox_rt/list_column_cpu.py
+++ b/torcharrow/velox_rt/list_column_cpu.py
@@ -6,7 +6,7 @@
 
 import array as ar
 import warnings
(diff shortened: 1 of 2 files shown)


torcharrow/velox_rt/list_column_cpu.py
```diff
@@ -6,7 +6,7 @@
 
 import array as ar
 import warnings
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Union
```
bearzx, line 9: what is Union here for?

bearzx (approved):
looks great to me

PR #363, opened 2022-06-03
Fix Torcharrow SSE Build

kgpai:
Velox sha : d7e4058a59255cc16d852d39bd927e779c63cfc4 has fix which ensure dependencies are built without avx flags which was causing torcharrow wheels to not run on Apple M1's.


wenleix (approved):
LGTM

PR #378, opened 2022-06-14
Add bison and flex in deps

wenleix:
Dependency added in Velox since https://github.com/facebookincubator/velox/commit/eb79237434026360ba3009e3c56b3b227ef6321c

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -16,9 +16,6 @@ jobs:
     runs-on: macos-latest
 
     steps:
-      - name: Print CPU info
-        run: sysctl -a | grep machdep.cpu
-
       - name: Setup Python environment
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/macos.yml
```diff
@@ -46,7 +43,7 @@ jobs:
 
       - name: Install dependencies from Brew
         run: |
-          brew install --formula ninja cmake ccache icu4c boost gflags glog libevent
+          brew install --formula ninja flex bison cmake ccache icu4c boost gflags glog libevent
```
kgpai, line 46: maybe move this to _setup-macos.sh ? This way we need make changes only there and not have to make changes to nightly-conda too.

kgpai (approved):

kgpai:
for macos its not required, for linux yes I can build a image or like here we can add the apt-get install command before building the binary..

.github/workflows/macos.yml
wenleix: Want to make sure we are using the same command in CI as we illustrates in the README about how to build from source: https://github.com/facebookresearch/torcharrow#install-dependencies

This helps CI to capture more potential issues for community developer experience.

bearzx (approved):

PR #380, opened 2022-06-14
Update version info for the builds

bearzx:
Per release runbook, we'll need to have a version.txt to specify the current version. This is used as the source of truth and we shall update it after every branch cut. This PR adds version.txt and does necessary adjustments for setup.py to use it.


ejguan (commented):
Do you also want to change here?
https://github.com/facebookresearch/torcharrow/blob/e086db7941bdc6ba8fe6e76d9f780022c40914fe/setup.py#L22

bearzx:
> Do you also want to change here?
> 
> https://github.com/facebookresearch/torcharrow/blob/e086db7941bdc6ba8fe6e76d9f780022c40914fe/setup.py#L22

good point, this one works a default value, so as long as we use the build_wheel.sh it will be overwritten as the one in version.txt. But it does make sense to also point this value to version.txt as well.

wenleix (approved):
Thanks!

PR #384, opened 2022-06-17
Create add_tokens UDF

Nayef211:
## Description
- Adding and registering UDF for `add_tokens` which can operate on list of `int` and `str`

# Testing
`pytest torcharrow/test/transformation/test_text_ops.py`

### DIFF
--- a/csrc/velox/functions/functions.h
+++ b/csrc/velox/functions/functions.h
@@ -16,6 +16,7 @@
 #include "rec/sigrid_hash.h" // @manual
 #include "string_functions.h"
 #ifdef USE_TORCH
+#include "text/add_tokens.h" // @manual
 #include "text/bpe_tokenize.h" // @manual
 #include "text/vocab_ops.h" // @manual
 #endif
@@ -262,6 +263,19 @@ inline void registerTorchArrowFunctions() {
       velox::ArrayWriterT<velox::Varchar>,
       std::shared_ptr<GPT2BPEEncoder>,
       velox::Varchar>({"bpe_tokenize"});
+
+  velox::registerFunction<
+      add_tokens,
+      velox::Array<int64_t>,
+      velox::Array<int64_t>,
+      velox::Array<int64_t>,
+      bool>({"add_tokens"});
+  velox::registerFunction<
+      add_tokens,
+      velox::Array<velox::Varchar>,
+      velox::Array<velox::Varchar>,
+      velox::Array<velox::Varchar>,
+      bool>({"add_tokens"});
 #endif
 
   //   sigrid_hash
--- a/csrc/velox/functions/text/add_tokens.h
+++ b/csrc/velox/functions/text/add_tokens.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include "velox/functions/Udf.h"
+#include "velox/type/Type.h"
+
+namespace facebook::torcharrow::functions {
+template <typename T>
+struct add_tokens {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  template <typename TInput, typename TOutput, typename TTokens>
+  FOLLY_ALWAYS_INLINE void call(
+      TOutput& output,
+      const TInput& input,
+      const TTokens& tokens,
+      bool begin = true) {
+    output.reserve(input.size() + tokens.size());
+    if (begin) {
+      output.add_items(tokens);
+      output.add_items(input);
+    } else {
+      output.add_items(input);
+      output.add_items(tokens);
+    }
+  }
+};
+} // namespace facebook::torcharrow::functions
--- a/torcharrow/functional.py
+++ b/torcharrow/functional.py
@@ -110,6 +110,35 @@ def __getattr__(op_name: str):
     return wrapper
 
 
+### operations in for text domain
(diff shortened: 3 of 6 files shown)


csrc/velox/functions/functions.h
```diff
@@ -262,6 +263,19 @@ inline void registerTorchArrowFunctions() {
       velox::ArrayWriterT<velox::Varchar>,
       std::shared_ptr<GPT2BPEEncoder>,
       velox::Varchar>({"bpe_tokenize"});
+
+    velox::registerFunction<
+      add_tokens,
+      velox::Array<int32_t>,
+      velox::Array<int32_t>,
+      velox::Array<int32_t>,
```
bearzx, line 271: maybe also try add one with int64_t? I remember python by default cast all the integer literals to int64.

torcharrow/test/transformation/test_text_ops.py
```diff
@@ -114,6 +116,19 @@ def setUpClass(cls):
...
+
+        cls.base_df_add_token = ta.dataframe(
+            {
+                "text": [["Hello", "world"], ["How", "are", "you!"]],
+                "indices": [[1, 2], [3, 4, 5]],
+            },
+            dtype=dt.Struct(
+                fields=[
+                    dt.Field("text", dt.List(dt.string)),
+                    dt.Field("indices", dt.List(dt.int64)),
```
wenleix, line 128: @bearzx  is right. `df["indices"]` is int64 but `add_tokens` expects `int32`, which cause type mismatch

cc @Nayef211

torcharrow/functional.py
```diff
@@ -104,6 +104,25 @@ def __getattr__(op_name: str):
...
     return wrapper
 
+### operations in for text domain
+def add_tokens(
+    input_col: Union[ListColumn, List[Union[int, str]]],
+    tokens: Union[ListColumn, List[Union[int, str]]],
+    begin: bool,
+) -> NumericalColumn:
+    """
+    Append or prepend a token to a column.
```
wenleix, line 114: thanks for adding the doc string!

torcharrow/test/transformation/test_text_ops.py
Nayef211: This did the trick thanks @bearzx and @wenleix!

torcharrow/functional.py
Nayef211: QQ: should we add each UDF we implement to the functional alongside documentation? I.e. would it be best practice for us to do this with `bpe_tokenize` and `lookup_indices` UDFs?

torcharrow/functional.py
```diff
@@ -109,6 +109,34 @@ def __getattr__(op_name: str):
...
     return wrapper
 
+### operations in for text domain
+def add_tokens(
+    input_col: Union[ListColumn, List[Union[int, str]]],
+    tokens: Union[ListColumn, List[Union[int, str]]],
+    begin: bool,
+) -> NumericalColumn:
+    """
+    Append or prepend a list of tokens/indices to a column.
```
wenleix, line 119: Thanks for the nice stub and doc string!

Mind if also add a entry into the functional doc? (e.g. `Text Operations`) : 

https://github.com/pytorch/torcharrow/blame/d20e01b0cbb6cad4be10526344422b0ef6ff4efd/docs/source/functional.rst#L35

wenleix (approved):

csrc/velox/functions/text/add_tokens.h
```diff
@@ -0,0 +1,41 @@
...
+
+        template <typename TInput, typename TOutput, typename TTokens>
+        FOLLY_ALWAYS_INLINE void call(
+            TOutput &output,
+            const TInput &input,
+            const TTokens &tokens,
+            bool begin = true)
+        {
+            output.reserve(input.size() + tokens.size());
+            if (begin)
```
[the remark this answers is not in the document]
parmeet: hmm, that's a good question but I don't think there is such use-case in text AFAIK. But this operator may be consumed beyond text I guess..So if there exist such use-case outside of text, we could potentially add this option :)

torcharrow/functional.py
```diff
@@ -109,6 +109,34 @@ def __getattr__(op_name: str):
...
+    begin: bool,
+) -> NumericalColumn:
+    """
+    Append or prepend a list of tokens/indices to a column. 
+    This is a common operation to add EOS and BOS tokens to text.
+
+    Parameters
+    ----------
+    input_col: Column that has a list of input tokens/indices
+    tokens: Column that has a list of tokens/indices to append or prepend
```
parmeet, line 125: should we just say `list of tokens/indices to append or prepend`?

Nayef211: Sure just updated the docstring!

torcharrow/functional.py
Nayef211: @wenleix lmk if you have any thoughts on this

torcharrow/functional.py
Nayef211: Yup done!

torcharrow/functional.py
wenleix: > should we add each UDF we implement to the functional alongside documentation?

Yes. But since `bpe_tokenize` and `lookup_indices` don't build by default in TorchArrow package, so we can defer adding it a bit.

csrc/velox/functions/text/add_tokens.h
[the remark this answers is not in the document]
wenleix: The benchmark[1] seems to always add two tokens? Perhaps we can have an overloaded version of `add_tokens` that add both beginning and the end to avoid one list allocation? 

[1] https://github.com/pytorch/text/pull/1807/files

PR #385, opened 2022-06-17
Doc update

bearzx:
Changes in this PR:
* Put doc build and deploy into a separate workflow, so it doesn't run every time with a push. It also uploads the docs to the beta folder, which is accordingly created and linked in gh-pages branch.
* Added template parameters in the .rst files as @NivekT suggested.
*  For the `istring_column.StringMethods` for now the best I could do is just removing the torcharrow prefix. I feel it's sort of related to the fact that StringMethods is not part of StringColumn. Importing it with an alias did not seem to work. We can keep investigating to find the best way to present it in the doc.

### DIFF
--- a/.github/workflows/deploy-doc.yml
+++ b/.github/workflows/deploy-doc.yml
@@ -1,0 +1,39 @@
+name: Deploy Doc
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Check out source repository
+        uses: actions/checkout@v2
+        with:
+            ref: "release/0.1.0"
+            submodules: recursive
+
+      - name: Install TorchArrow
+        run: |
+          pip install --pre torcharrow -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+
+      - name: Build the docs
+        run: |
+          cd ./docs
+          pip install -r requirements.txt --user
+          make html
+          cd ..
+
+      - name: Deploy Docs on Push
+        uses: JamesIves/github-pages-deploy-action@v4.2.5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: gh-pages # The branch the action should deploy to.
+          folder: docs/build/html # The folder the action should deploy.
+          target-folder: 0.1.0
--- a/docs/source/column.rst
+++ b/docs/source/column.rst
@@ -3,8 +3,8 @@
 torcharrow.Column
 ==========================
 
-A :class:`torcharrow.Column` is a 1-dimension torch.Tensor like data structure containing 
-elements of a single data type. It also supports non-numeric types such as string, 
+A :class:`torcharrow.Column` is a 1-dimension torch.Tensor like data structure containing
+elements of a single data type. It also supports non-numeric types such as string,
 list, struct.
 
 Data types
@@ -40,6 +40,7 @@ Column class reference
 .. autosummary::
     :toctree: generated
     :nosignatures:
+    :template: class.rst
 
     Column.head
     Column.tail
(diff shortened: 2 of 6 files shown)


.github/workflows/deploy-doc.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Check out source repository
+        uses: actions/checkout@v2
+        with:
+            # ref: "release/0.1.0"
+            submodules: recursive
+
+      - name: Install TorchArrow
+        run: |
+          pip install --pre torcharrow -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
```
wenleix, line 27: Does the schedule make sure we can get the nightly build on the same day? ^_^

docs/source/dataframe.rst
```diff
@@ -25,6 +25,7 @@ DataFrame Class and General APIs
 .. autosummary::
     :toctree: generated
     :nosignatures:
+    :template: class.rst
```
wenleix, line 28: I copied the `class.rst` and `fucntion.rst` from TorchData or some other domain library, but I actually don't know what style change they bring 😳

wenleix (approved):
LGTM. Thanks!

.github/workflows/deploy-doc.yml
bearzx: I think there's no guarantee --- depending on which one runs first. I'm actually thinking to change this one to manual run only, since it writes to the "release" folder and shouldn't be needed to run nightly. But in this case, we'll need to be careful which version of torcharrow build it installs. Here I'm using the nightly build because that's the one available, but ideally I should use a RC build, which is currently not uploaded anywhere.

On the other hand, we can probably leave a build and deploy doc section to one of the nightly builds, and make sure it uploads the doc to the "main" folder (like [torchdata](https://pytorch.org/data/versions.html)).

But yeah this still needs some tweaks.

docs/source/dataframe.rst
bearzx: me either ... but seems no harm to leave it here.

docs/source/column.rst
```diff
@@ -63,7 +64,7 @@ Column class reference
     Column.to_arrow
     Column.to_tensor
     Column.to_pylist
-    Column.to_pandas    
+    Column.to_pandas
```
wenleix, line 67: separate to this PR -- looks like the current Pandas conversion is not efficient : 

https://github.com/pytorch/torcharrow/blob/main/torcharrow/icolumn.py#L1104-L1106

Maybe we should use something like `self.to_arrow().to_pandas()` to fix it? :)

docs/source/column.rst
```diff
@@ -96,42 +98,44 @@ StringColumn class reference
 .. autosummary::
     :toctree: generated
     :nosignatures:
+    :template: class.rst
+
+    istring_column.StringMethods.length
```
wenleix, line 103: as a follow-up, maybe we should rename `istring_column.py` to `string_column.py` to get better doc :) 

(btw, how does Pandas render doc or implement `str_col.str.XXX`...)

bearzx: is this how pandas implemented it? https://github.com/pandas-dev/pandas/blob/main/pandas/core/series.py#L5851

docs/source/column.rst
```diff
@@ -40,6 +40,7 @@ Column class reference
 .. autosummary::
     :toctree: generated
     :nosignatures:
+    :template: class.rst
```
NivekT, line 43: You may want to check if the output page looks as you expect.

bearzx:
> Some changes to the `gh-pages` branch is needed to make this work properly.
> 
> 1. We need a folder `main` where the nightly version (or just commits of `main` for now) should deploy to. You can move all the content there (except the `0.1.0` folder) into that folder.
> 2. The `0.1.0` branch should track changes from the release branch.
> 3. Add a [HTML file](https://github.com/pytorch/data/blob/gh-pages/index.html) that redirects to `beta` and a [list of versions](https://github.com/pytorch/data/blob/gh-pages/versions.html).
> 4. Add a symlink from `beta` to `0.1.0`.

yep gonna do that next ... I actually added some last Friday but then they were overwritten by the nightly doc deploy :D

docs/source/column.rst
bearzx: yep it looks the same to me with or without haha.

PR #387, opened 2022-06-20
Update RC build CI workflow

bearzx:
In this PR I made the following changes:
* added python 3.10 build
* added the pypi uploading part

One thing to note is, pypi doesn't allow uploading with the same name, but since we only manually trigger this workflow, we can always manually check if the files already exists and remove them if necessary. Another idea is to use TestPyPI (https://packaging.python.org/en/latest/guides/using-testpypi/).


ejguan (commented):
For the problem uploading multiple times, you can use `upload-artifact` and `download-artifact` to combine all binaries together then upload them.
See workflow in TorchData: https://github.com/pytorch/data/blob/32e447f1890437fa9707bc2e517cfa8b3e19dfca/.github/workflows/_build_test_upload.yml#L161-L179

bearzx:
> For the problem uploading multiple times, you can use `upload-artifact` and `download-artifact` to combine all binaries together then upload them. See workflow in TorchData: https://github.com/pytorch/data/blob/32e447f1890437fa9707bc2e517cfa8b3e19dfca/.github/workflows/_build_test_upload.yml#L161-L179

How does it solve the problem of uploading the same file to pypi? I thought if you already have file A on pypi and trying to uploading file A (exactly the same name) again pypi will return a "file exists" error.

ejguan:
> > For the problem uploading multiple times, you can use `upload-artifact` and `download-artifact` to combine all binaries together then upload them. See workflow in TorchData: https://github.com/pytorch/data/blob/32e447f1890437fa9707bc2e517cfa8b3e19dfca/.github/workflows/_build_test_upload.yml#L161-L179
> 
> How does it solve the problem of uploading the same file to pypi? I thought if you already have file A on pypi and trying to uploading file A (exactly the same name) again pypi will return a "file exists" error.

Wait. Why the files are the same. They should have different extensions, right?

PR #395, opened 2022-06-22
Minor changes to setup scripts

wenleix:
* Update folly version to be consistent with Velox: https://github.com/facebookincubator/velox/blob/c4237370e40fce925a5c34bd805008248d42684f/scripts/setup-ubuntu.sh#L25
* Minor config flag change (e.g. disable testing), following https://github.com/facebookincubator/velox/blob/04a85f7cb7fcc2eaa3b30b1749c3d97b3bba8676/scripts/setup-macos.sh#L104
and  https://github.com/facebookincubator/velox/blob/04a85f7cb7fcc2eaa3b30b1749c3d97b3bba8676/scripts/setup-macos.sh#L109

### DIFF
--- a/scripts/_setup-macos.sh
+++ b/scripts/_setup-macos.sh
@@ -27,7 +27,7 @@ set -x # Print commands that are executed.
 SCRIPTDIR=$(dirname "${BASH_SOURCE[0]}")
 source $SCRIPTDIR/../csrc/velox/velox/scripts/setup-helper-functions.sh
 
-FB_OS_VERSION=v2021.05.10.00
+FB_OS_VERSION=v2022.03.14.00
 NPROC=$(sysctl -n hw.physicalcpu)
 COMPILER_FLAGS=$(get_cxx_flags $CPU_TARGET)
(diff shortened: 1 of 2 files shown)


bearzx (approved):
LGTM. In the future, should we use the set up scripts from velox folder directly?

wenleix:
@bearzx : agree. One thing need to figure out is I would still like the developer to be able to execute `scripts/setup-xxx.sh` in https://github.com/pytorch/torcharrow#install-dependencies , rather than using `csrc/velox/velox/velox/scripts/...`. Maybe we can let `scripts/setup-xxx.sh` to delegate to `csrc/velox/velox/velox/scripts/...` ? :) 

We also need to confirm with Velox team that these scripts will exist and supported (otherwise, if some day Velox repo rename / removed the script, TorchArrow build would also fail). cc @kgpai

kgpai:
@wenleix Yes, they will be supported . They are also used for presto_cpp so no reason for TA not to rely on them.

PR #401, opened 2022-06-24
Fix performance issues in gpt2_bpe_tokenizer

laithsakka:
Summary: complex structures in c++ should be passed as const ref instead of value to avoid data copy. A bunch of functions was passing by value gpt2_bpe_tokenizer

Differential Revision: D37423480

### DIFF
--- a/csrc/velox/functions/text/gpt2_bpe_tokenizer.cpp
+++ b/csrc/velox/functions/text/gpt2_bpe_tokenizer.cpp
@@ -34,7 +34,8 @@ bool is_whitespace(const std::string& input) {
 }
 
 template <class Key_, class Value_>
-c10::Dict<Key_, Value_> _map_to_c10_dict(std::unordered_map<Key_, Value_> m) {
+c10::Dict<Key_, Value_> _map_to_c10_dict(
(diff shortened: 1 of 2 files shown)


wenleix (approved):
Wow.Thanks! 

AFAIK these functions are copied from TorchText.

Does that mean a lot of functions in TorchText itself (e.g. https://github.com/pytorch/text/blob/88b251f9cebae86feb5edf459b978bf211b65183/torchtext/csrc/gpt2_bpe_tokenizer.cpp#L120 ) can also be optimized? 

cc @parmeet , @Nayef211

Nayef211:
> Wow.Thanks!
> 
> AFAIK these functions are copied from TorchText.
> 
> Does that mean a lot of functions in TorchText itself (e.g. https://github.com/pytorch/text/blob/88b251f9cebae86feb5edf459b978bf211b65183/torchtext/csrc/gpt2_bpe_tokenizer.cpp#L120 ) can also be optimized?

Yup thanks for catching this @laithsakka. We can create a followup PR in torchtext with these changes.

cc @abhinavarora

Nayef211 (approved):

Issue #404, opened 2022-06-26
Supports `df.drop("a")`


wenleix:
Pandas's drop doesn't directly support it (and has to do `df.drop(["a", "b"])`:

```python
>>> import pandas as pd
>>> df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
>>> df.drop("a")
  File "<stdin>", line 1, in <module>
  File "/Users/wxie/opt/miniconda3/lib/python3.9/site-packages/pandas-1.3.4-py3.9-macosx-10.9-x86_64.egg/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/Users/wxie/opt/miniconda3/lib/python3.9/site-packages/pandas-1.3.4-py3.9-macosx-10.9-x86_64.egg/pandas/core/frame.py", line 4906, in drop
    return super().drop(
  File "/Users/wxie/opt/miniconda3/lib/python3.9/site-packages/pandas-1.3.4-py3.9-macosx-10.9-x86_64.egg/pandas/core/generic.py", line 4150, in drop
    obj = obj._drop_axis(labels, axis, level=level, errors=errors)
  File "/Users/wxie/opt/miniconda3/lib/python3.9/site-packages/pandas-1.3.4-py3.9-macosx-10.9-x86_64.egg/pandas/core/generic.py", line 4185, in _drop_axis
    new_axis = axis.drop(labels, errors=errors)
  File "/Users/wxie/opt/miniconda3/lib/python3.9/site-packages/pandas-1.3.4-py3.9-macosx-10.9-x86_64.egg/pandas/core/indexes/base.py", line 6017, in drop
    raise KeyError(f"{labels[mask]} not found in axis")
KeyError: "['a'] not found in axis"
```

However, Pandas does support "label" as single parameter or list. The error is because in Pandas, both "index" and "column" axis are supported (and "index" axis can also be string-named) . And by default the `axis=0` makes it error (no "index" with name "a"). The following would work:

```python
>>> df.drop("a", axis=1)
   b
0  4
1  5
2  6
```

TorchArrow in general designed to not support complicated index-manipulation in most cases (index is always number, and we only support thing like `slice` and `append`). And column is the default operation. So there is no ambiguity to support `df.drop("a")`

wenleix:
cc @dracifer

Krish-bhardwaj:
[...]

PR #419, opened 2022-07-01
add proper check _check_columns

waitingkuo:
There're several functions that use _check_columns to check whether the input sequence of str are in dataframe's columns

```python
    def _check_columns(self, columns: Iterable[str]):
          valid_names = {f.name for f in self.dtype.fields}
          for n in columns:
              if n not in valid_names:
                  raise TypeError(f"column {n} not among existing dataframe columns")
```
  
If we input the columns as a str, this function will loop the string and verify whether each character is in the columns or not.

e.g.

```python
     import torcharrow as ta

     df = ta.dataframe(
        {"aa": list(range(7)), 
          "b": list(reversed(range(7))), 
           "c": list(range(7))
         })
    
     df.groupby('aa')
```

this raise 

      TypeError: column a not among existing dataframe columns

But df.groupby('bb') didn't

this pr is to raise exception is the input itself is a string

### DIFF
--- a/torcharrow/velox_rt/dataframe_cpu.py
+++ b/torcharrow/velox_rt/dataframe_cpu.py
@@ -308,6 +308,9 @@ class DataFrameCpu(ColumnCpuMixin, DataFrame):
             return self
 
     def _check_columns(self, columns: Iterable[str]):
+        if isinstance(columns, str):
+            raise TypeError(f"columns should be Iterable of str but not str")
+
         valid_names = {f.name for f in self.dtype.fields}
         for n in columns:
             if n not in valid_names:
@@ -1597,11 +1600,13 @@ class DataFrameCpu(ColumnCpuMixin, DataFrame):
     @expression
     def drop_duplicates(
         self,
-        subset: Optional[List[str]] = None,
+        subset: Optional[Union[str, List[str]]] = None,
         keep="first",
     ):
         self._prototype_support_warning("drop_duplicates")
 
+        if isinstance(subset, str):
+            subset = [subset]
         columns = subset if subset is not None else self.columns
         self._check_columns(columns)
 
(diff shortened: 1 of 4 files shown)


torcharrow/velox_rt/dataframe_cpu.py
```diff
@@ -308,6 +308,9 @@ def append(self, values: Iterable[Union[None, dict, tuple]]):
             return self
 
     def _check_columns(self, columns: Iterable[str]):
+        if isinstance(columns, str):
```
wenleix, line 311: should we just check `columns` should be a `Iterable[str]`? -- basically enforce the type annotation?

wenleix (approved):
LGTM. Perhaps also add a test for the behavior? 

Also see discussions in https://github.com/pytorch/torcharrow/issues/404#issuecomment-1175210519 about `_check_columns` and `drop`.

torcharrow/velox_rt/dataframe_cpu.py
waitingkuo: any advice to check it? ```isinstance(columns, Iterable[str])``` doesnt work

wenleix:
Ah I see. For `groupby`, it makes sense to support single string, similar to Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html. So I guess it's the option 2 in your proposal :)

waitingkuo:
fixing _check_columns trigger some failed test cases for  ```drop```, ````groupby```, and ```drop_duplicates```. i extended them to accept single str as column/subset 

#404 is enabled in this pr

wenleix (approved):
LGTM. Thanks for the contribution! Would you mind squash the commits into one before I merge? (such as `Enable single column str for drop, drop_duplicates and groupby`? )

waitingkuo:
@wenleix  squashed

wenleix:
Merged #419 as https://github.com/pytorch/torcharrow/commit/a700ae1da664c82e1abdca63dfa7eeae05952ae9 . Thanks for the contribution!

PR #423, opened 2022-07-05
Bump versions for 0.1.0 beta release

bearzx:
As titled, this is to leave a more formal record in the system that we are ready to start releasing.


wenleix (approved):
LGTM. Make sure squash the commits locally and do a force push, so there is only one commit (and no merge)

PR #431, opened 2022-07-12
Fix breaking TA ubuntu build

kgpai:
Uses new Velox image that uses openssl 1.1 . Tested locally.


kgpai:
FYI, The builds succeed, but writes to s3 fail (which might be expected).

wenleix (approved):
LGTM. 

Can you sqaush all the commits together before I merge it? -- TorchArrow moved to "GitHub first" model so we cannot rely on Phabricator to squash everything now :)

PR #446, opened 2022-07-15
add doc deploy to nightly ubuntu wheel build

bearzx:
tsia. We previously removed the doc deploy from "CI on push" because we didn't want to update the doc on every PR push, this PR recreates the doc push in nightly build workflows.

Did some minor cleanup as well.

### DIFF
--- a/.github/workflows/nightly-wheel.yml
+++ b/.github/workflows/nightly-wheel.yml
@@ -9,6 +9,9 @@ on:
       AWS_PYTORCH_UPLOADER_SECRET_ACCESS_KEY:
         required: true
 
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
 jobs:
   linux-container:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-wheel.yml
```diff
@@ -56,6 +53,25 @@ jobs:
               ~/.local/bin/aws s3 cp "$pkg" "$S3_PATH" --acl public-read
           done
 
+      - name: Install the freshly uploaded TorchArrow
+        run: |
+          pip install --pre torcharrow -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
```
wenleix, line 58: curious: does this guarantee to download the just uploaded TA nightly? ;) 

cc @ejguan , @NivekT

wenleix (approved):
Thanks!

.github/workflows/nightly-wheel.yml
ejguan: Probably not. I believe it will take a few minutes for the just uploaded wheels becoming available from the link. Why not just install from your local wheel?

bearzx: yep that's probably better, lemme tweak it that way.

PR #455, opened 2022-07-21
Added all text ops to functional

Nayef211:
## Description
- Added all text ops to `functional.py` along with doc strings showcasing usage

## Testing
`pytest torcharrow/test/transformation/test_text_ops.py`

### DIFF
--- a/torcharrow/functional.py
+++ b/torcharrow/functional.py
@@ -4,12 +4,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from msilib.schema import Error
 import sys
 from functools import partial
 from types import ModuleType
 from typing import Dict, List, Optional, Set, Union
 
 import torcharrow.dtypes as dt
+import torcharrow._torcharrow as _ta
 
 from torcharrow.icolumn import Column
 from torcharrow.ilist_column import ListColumn
@@ -121,6 +123,7 @@ def add_tokens(
     """
     Append or prepend a list of tokens/indices to a column.
     This is a common operation to add EOS and BOS tokens to text.
+    Note: this method is only available if torcharrow is built from source with the `USE_TORCH` flag enabled.
 
     Parameters
     ----------
@@ -138,6 +141,10 @@ def add_tokens(
     1  [0, 3, 4, 5]
     dtype: List(Int64(nullable=True), nullable=True), length: 2, null_count: 0
     """
+    if not _ta.is_built_with_torch():
+        raise Exception(
+            "This functionality requires TorchArrow to be built with PyTorch. Please install the library from source with the `USE_TORCH` flag enabled."
+        )
     return _dispatch("add_tokens", input_col, tokens, begin)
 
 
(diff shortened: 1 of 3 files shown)


torcharrow/functional.py
```diff
@@ -141,6 +141,60 @@ def add_tokens(
     return _dispatch("add_tokens", input_col, tokens, begin)
 
 
+def bpe_tokenize(
+    tokenizer,
```
Nayef211, line 145: @wenleix I'm wondering what the correct type would be for the `tokenizer` param here

torcharrow/functional.py
```diff
@@ -141,6 +141,60 @@ def add_tokens(
...
+    >>> functional.bpe_tokenize(tokenizer, a)
+    0  ['15496', '2159', '28265', '703', '389', '345', '30']
+    1  ['4965', '11377', '64', '2208', '72', '29625']
+    dtype: List(String(nullable=True), nullable=True), length: 2, null_count: 0
+    """
+    return _dispatch("bpe_tokenize", tokenizer, input_col)
+
+
+def lookup_indices(
+    vocab,
```
Nayef211, line 172: @wenleix same question for `vocab` param

torcharrow/functional.py
```diff
@@ -141,6 +141,60 @@ def add_tokens(
...
+    Parameters
+    ----------
+    tokenizer: An instance of the `GPT2BPEEncoder` class
+    input_col: List of input text
+
+    Examples
+    --------
+    >>> import torcharrow as ta
+    >>> from torcharrow import functional
+    >>> # init_bpe_encoder() is helper function that creates an instance of the `GPT2BPEEncoder` class
```
parmeet, line 160: should we show instantiation of _ta.GPT2BPEEncoder like you did for _ta.Vocab? Or alternatively give reference to init_bpe_encoder function from here https://github.com/pytorch/torcharrow/blob/cecbd9222dca54784b21c6e914869c90b0fec60e/torcharrow/test/transformation/test_text_ops.py#L59?

torcharrow/functional.py
```diff
@@ -141,6 +141,60 @@ def add_tokens(
     return _dispatch("add_tokens", input_col, tokens, begin)
 
 
+def bpe_tokenize(
+    tokenizer,
+    input_col: Union[ListColumn, List[str]],
+) -> ListColumn:
+    """
+    Tokenize text into a list of tokens using the GPT-2 BPE Tokenizer
```
wenleix, line 149: I think we need to make clearly in docstring that this method is only useable if TA is built with `USE_TORCH=1`?

torcharrow/functional.py
```diff
@@ -141,6 +141,60 @@ def add_tokens(
...
+    >>> from torcharrow import functional
+    >>> # init_bpe_encoder() is helper function that creates an instance of the `GPT2BPEEncoder` class
+    >>> tokenizer = init_bpe_encoder()
+    >>> a = ta.column([["Hello World!, how are you?", "Respublica superiorem"]])
+    >>> functional.bpe_tokenize(tokenizer, a)
+    0  ['15496', '2159', '28265', '703', '389', '345', '30']
+    1  ['4965', '11377', '64', '2208', '72', '29625']
+    dtype: List(String(nullable=True), nullable=True), length: 2, null_count: 0
+    """
+    return _dispatch("bpe_tokenize", tokenizer, input_col)
```
wenleix, line 168: can check`_ta.is_built_with_torch()` and raise a more explicit error message? (otherwise, it will just say "`bpe_tokenize` function not found~)

wenleix:
Thanks! These docs are nice!

parmeet (approved):
I think overall LGTM! Let's just address the comments before merging.

Nayef211:
@wenleix @parmeet thanks for the feedback! I should have resolve your PR comments. @wenleix just wanted to see if you could give the changes one more pass before I merge this in 😄

torcharrow/functional.py
```diff
@@ -121,6 +123,7 @@ def add_tokens(
     """
     Append or prepend a list of tokens/indices to a column.
     This is a common operation to add EOS and BOS tokens to text.
+    Note: this method is only available if torcharrow is built from source with the `USE_TORCH` flag enabled.
```
wenleix, line 126: Isn't this method in general available? (it doesn't require any thing from libtorch)

wenleix (approved):
LGTM

torcharrow/functional.py
```diff
@@ -4,12 +4,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from msilib.schema import Error
```
wenleix, line 7: This looks not needed? (And cause test failure )

wenleix:
Also make sure to squash commits into one before merge :)

torcharrow/functional.py
Nayef211: Yup just removed!

torcharrow/functional.py
Nayef211: Yup you're right. I moved the corresponding UDF out of the `ifdef` statement so that this method is available regardless of if TA is built with torch. Thanks for the catch!

bearzx:
> Also make sure to squash commits into one before merge :)

I enabled the "squash and merge" option on github so hopefully we don't need to run squash commands anymore :D

PR #468, opened 2022-08-02
Refactor DType definition into dtypes_core.py

wenleix:
Summary:
dtypes.py contains of two parts:
(1) Standard Arrow-compatible DataType definition (Int8/16/32/64, Float32/64, String, List, Map, Struct)
(2) Utility functions around DType

The DType definition is quite standalone and stable now; Refactor DataType definition into seperate file so allows reuse in next gen TorchArrow, such as TorchArrow-UPM or Tensor-based TA

In theory we could let dtypes.py only contain DType definition, and move other things into `dtype_util.py`. Starting with this first step.

Differential Revision: D38358389

### DIFF
--- a/torcharrow/dtypes_core.py
+++ b/torcharrow/dtypes_core.py
@@ -1,0 +1,342 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import dataclasses
+import re
+import typing as ty
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+# -----------------------------------------------------------------------------
+# Aux
+
+# Pretty printing constants; reused everywhere
+# TODO: Inline these constants.
+OPEN = "{"
+CLOSE = "}"
+NL = "\n"
+
+
+# -----------------------------------------------------------------------------
+# Schema and Field
+
+MetaData = ty.Dict[str, str]
+
+
+@dataclass(frozen=True)
+class Field:
+    name: str
+    dtype: "DType"
+    metadata: ty.Optional[MetaData] = None
+
+    def __str__(self):
+        meta = ""
+        if self.metadata is not None:
+            meta = (
+                f"meta = {OPEN}{', '.join(f'{k}: {v}' for k,v in self.metadata)}{CLOSE}"
+            )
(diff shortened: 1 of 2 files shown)


torcharrow/dtypes_core.py
```diff
@@ -0,0 +1,342 @@
...
+        nullable = ", nullable=" + str(self.nullable) if self.nullable else ""
+        return f"Map({self.key_dtype}, {self.item_dtype}{nullable})"
+
+    def default_value(self):
+        return {}
+
+
+@dataclass(frozen=True)
+class List(DType):
+    item_dtype: DType
```
wenleix, line 216: Here we use `item_dtype` and `fixed_size` while in PyArrow it's called `value_type` and `list_size`: https://arrow.apache.org/docs/python/generated/pyarrow.list_.html

 `fixed_size` vs. `list_size`  seems fine to me.  I guess `item` can be renamed to `value` :). (since `ListColumn` also contains `values` and `offsets`.

torcharrow/dtypes_core.py
```diff
@@ -0,0 +1,342 @@
...
+    default: ty.ClassVar[str] = ""
+
+    def constructor(self, nullable):
+        return String(nullable)
+
+
+@dataclass(frozen=True)
+class Map(DType):
+    key_dtype: DType
+    item_dtype: DType
```
wenleix, line 192: I think this `key/item` is coming from PyArrow's Map Type: https://arrow.apache.org/docs/python/generated/pyarrow.map_.html

But in many columnar Database (Presto/Velox) calls it `key/value` instead. `key/value` also seems to align with Pythonic API (in Python Dict, `item` means the (key, value) pair). 


Velox: https://github.com/facebookincubator/velox/blob/2148956b208132e4e37684392d28651e5ef9ff07/velox/vector/ComplexVector.h#L401-L402

torcharrow/dtypes_core.py
wenleix: or, `element_dtype` as per `Velox` and `Trino/Presto`: 

Velox: https://github.com/facebookincubator/velox/blob/ee631f54e9646e0442b577926d6bff4633e93401/velox/type/Type.cpp#L577-L580

Trino: https://github.com/trinodb/trino/blob/8d6b82e874ea7e79722bccacc0ebbca950b8908f/core/trino-spi/src/main/java/io/trino/spi/type/ArrayType.java#L86-L90

PR #475, opened 2022-08-17
Reword README

wenleix:
* Emphasize it's `torch.Tensor`-like. We found this quite important for
ML developers to author model preproc. (words copied back from
https://github.com/pytorch/torcharrow/commit/f7f07c983cdce218d375995c4d5fd640eb4fd78d)
* Re-org features to emphasize the two high-level product spec:
(1) Strong integration with PyTorch and GPU (2) Strong integration with
Arrow and variable-width/nested type support. The implementation detail
can change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,14 @@
-# TorchArrow
+# TorchArrow: a data processing library for PyTorch
 
-**This library is currently in the Beta stage and does not have a stable release. The API may change based on
-user feedback or performance. We are committed to bring this library to stable release, but future changes may not be
-completely backward compatible. If you have suggestions on the API or use cases you'd like to be covered, please open a
+**This library is currently in the Beta stage and does not have a stable release. The API and implementation may change based on
+user feedback or performance. Future changes may not be backward compatible.
+If you have suggestions on the API or use cases you'd like to be covered, please open a
 GitHub issue. We'd love to hear thoughts and feedback.**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,18 +1,14 @@
...
+TorchArrow is a [torch](https://github.com/pytorch/pytorch).Tensor-like Python DataFrame library for data preprocessing in PyTorch models, with two high-level features:
 
-* Seamless handoff with [PyTorch](https://github.com/pytorch/pytorch) or other model authoring, such as Tensor collation and easily plugging into PyTorch DataLoader and [DataPipes](https://github.com/pytorch/data#what-are-datapipes)
-* Zero copy for external readers via [Arrow](https://github.com/apache/arrow) in-memory columnar format
-* Multiple execution runtimes support:
-    - High-performance CPU backend via [Velox](https://github.com/facebookincubator/velox/)
-    - (Future Work) GPU backend via [libcudf](https://docs.rapids.ai/api/libcudf/stable/)
-* High-performance C++ UDF support with vectorization
+* DataFrame library (like Pandas) with strong GPU acceleration (under development) and PyTorch ecosystem integration.
```
ejguan, line 10: Instead of only talking about GPU acceleration, I feel like it's better to say something like compiler and hardware optimization. The benefit of Tensor is not just GPU but also Jittable optimization on any specific backend hardware

ejguan (approved):

PR #479, opened 2022-08-24
Pass not nullable context and results by reference (#2367)

laithsakka:
Summary:
X-link: https://github.com/facebookincubator/velox/pull/2367

Those are always assumed not null and de-referenced with out
checking, hence passing them by reference is more suitable.

Differential Revision: D38946493


wenleix:
Landed into fbsync as https://github.com/pytorch/torcharrow/commit/458dc0bcbc778b6957afb08248c86f2a5ceaead1 and cherry-pick into main branch as https://github.com/pytorch/torcharrow/pull/484

